PR #308, opened 2015-01-22
run through supervisord (bug 1122190)

andymckay:


muffinresearch:
Doesn't match the conf location here: https://github.com/mozilla/marketplace-env/pull/60

r+wc

Dockerfile
```diff
@@ -4,6 +4,8 @@
 # NOTE: this is not provided for production usage.
 FROM mozillamarketplace/centos-mysql-mkt:0.2
 
+RUN yum install -y supervisor
```
muffinresearch: We should add this to that ^ base image at some point.

PR #309, opened 2015-01-23
run solitude as a proxy (bug 1112672)

andymckay:
* we only need to expose 2602, not the proxy
* move the environment variables to be per solitude instance, so inside supervisor.conf, not the dockerfile

### DIFF
--- a/bin/docker/supervisor.conf
+++ b/bin/docker/supervisor.conf
@@ -10,3 +10,22 @@ redirect_stderr=true
 stdout_logfile=logs/docker.log
 stdout_logfile_maxbytes=1MB
 stopsignal=KILL
+environment=
+    SOLITUDE_ZIPPY_PROXY="http://solitude:2603/proxy/provider",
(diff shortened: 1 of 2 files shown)


bin/docker/supervisor.conf
```diff
@@ -10,3 +10,18 @@ redirect_stderr=true
 stdout_logfile=logs/docker.log
 stdout_logfile_maxbytes=1MB
 stopsignal=KILL
+environment=SOLITUDE_ZIPPY_PROXY="http://solitude:2603/proxy/provider", SOLITUDE_DATABASE="mysql://root:@mysql:3306/solitude", MEMCACHE_URL="memcached:11211", SOLITUDE_URL="http://solitude:2602"
```
muffinresearch: Can this be wrapped without breaking the conf?

muffinresearch:
Shiny lgtm, r+wc

bin/docker/supervisor.conf
andymckay: they can indeed, testing it still works :)

PR #310, opened 2015-01-27
Activate Boku on stage

kumar303:


solitude/settings/sites/stage/proxy.py
```diff
@@ -52,7 +52,14 @@
...
 BANGO_AUTH = private.BANGO_AUTH
+
+ZIPPY_CONFIGURATION = {
+    'boku': {
+        'url': base.BOKU_API_DOMAIN,
+    }
+}
+
+BOKU_MERCHANT_ID = private.BOKU_MERCHANT_ID
+BOKU_SECRET_KEY = private.BOKU_SECRET_KEY
```
kumar303: note to self: I need to get ops to put the right values in here before merging. Probably something from https://mana.mozilla.org/wiki/display/MARKET/Boku

PR #311, opened 2015-01-30
Log some price info for Boku

kumar303:


andymckay:
r+, logging never hurts

PR #313, opened 2015-01-30
Boku price row-refs are 0-based (bug 1128012)

kumar303:


lib/boku/tests/test_client.py
```diff
@@ -1,3 +1,4 @@
+from decimal import Decimal
 import time
 import urlparse
```
andymckay: i think import before from?

andymckay:
Kinda surprised this happened, I swear we got to testing prices and payments on payments-alt, so I'm confused why it worked there. Or did it ever really work?

lib/boku/tests/test_client.py
kumar303: oh duh, forgot

andymckay:
r+ wc

PR #314, opened 2015-01-31
Pep8

andymckay:


lib/bango/resources/package.py
```diff
@@ -48,8 +48,11 @@ def build_bundle(self, obj=None, data=None, request=None):
     def dehydrate_full(self, bundle):
         if getattr(bundle, 'full', False):
             return response_to_dict(
-                self.client('GetPackage',
-                    {'packageId': bundle.obj.package_id}))
+                self.client(
+                    'GetPackage',
+                    {'packageId': bundle.obj.package_id}
```
kumar303: could probably be on one line

lib/bango/tests/test_forms.py
```diff
@@ -90,7 +90,7 @@ class TestPackage(APITest):
     def test_no_auth(self, clean):
         form = PackageForm(good_address)
         ok_(form.is_valid())
-        ok_(not 'eventNotificationURL' in form.bango_data)
+        ok_('eventNotificationURL' not in form.bango_data)
```
kumar303: hmm. I wonder if this will break (or fix?) anything.

kumar303: oh it's just a test

kumar303:
r+

lib/bango/tests/test_forms.py
andymckay: Both work, but the latter is preferred. I fixed a few of these in zamboni and the tests all passed.

PR #316, opened 2015-02-05
Bump docs reqs

muffinresearch:
Bump docs reqs to match webpay. This may stop some of the build errors from rtd.


kumar303:
oh sweet, does this fix RTD?

r+

PR #317, opened 2015-02-06
Send actual prices not refs (bug 1130119)

kumar303:


andymckay:
r+, do you need to change webpay when it calls solitude to pass through anything else, eg currency?

kumar303:
oh yeah, I forgot to make a pull for webpay. heh. It's patched locally.

PR #318, opened 2015-02-07
add in md5 sig check (bug 1130213)

andymckay:


lib/boku/client.py
```diff
@@ -282,6 +290,20 @@ def check_transaction(self, transaction_id):
...
+    def check_sig(self, data):
+        """
+        Check that a signature is valid. If solitude is run with the proxy,
+        this check has to be done on the proxy, because the solitude database
+        does not have access to the required BOKU_SECRET_KEY.
+
+        This takes the output of `get_pairs_string` as data.
+
+        Returns the value of the signature calculated using BOKU_SECRET_KEY.
+        """
```
kumar303: Because of your comment about how this needs to run on the proxy I wonder if it should raise an error if ``settings.BOKU_SECRET_KEY`` is empty. Otherwise, it will silently create the wrong signature (probably using an empty string as the secret)

lib/boku/client.py
```diff
@@ -282,6 +290,20 @@ def check_transaction(self, transaction_id):
             'paid': Decimal(tree.find('paid').text),
         }
 
+    def check_sig(self, data):
```
kumar303: I think ``get_sig()`` would be a better name for this function. Also, I think the docstring is wrong about it checking the signature. It looks like ``utils.check_sig()`` does that now, right?

lib/boku/tests/utils.py
```diff
@@ -71,14 +73,16 @@ def add_seller_boku(self):
...
+        data = {
             'action': 'billingresult',
             'amount': '100',
             'currency': 'MXN',
             'param': 'some:uuid',
-            'sig': 'some:sig',
             'trx-id': 'some:trxid'
         }
+        data['sig'] = get_boku_request_signature(
+            settings.BOKU_SECRET_KEY, get_pairs_string(data))
```
kumar303: I think ``solitude/settings/test.py`` should set a value for ``BOKU_SECRET_KEY`` for this.

lib/boku/utils.py
```diff
@@ -42,3 +42,19 @@ def error(msg):
         error('did not verify: {0} != {1}'.format(real, amount))
 
     log.info('transaction: {0} verified successfully'.format(transaction))
+
+
+def check_sig(data):
+    """
+    Check the signature in the request.
```
kumar303: it looks like the docstring from ``client.check_sig()`` should move over here

lib/boku/utils.py
```diff
@@ -42,3 +42,19 @@ def error(msg):
...
+    """
+    raw = data.copy()
+    boku_sig = raw.pop('sig')
+
+    client = get_client(settings.BOKU_MERCHANT_ID, settings.BOKU_SECRET_KEY)
+    our_sig = client.check_sig(get_pairs_string(raw))['sig']
+
+    if our_sig != boku_sig:
+        msg = 'signature check failed'
```
kumar303: is there an identifier we can log here too? Something like 

    'signature check failed for {t}'.format(t=data.get('transaction_id', '<unknown transaction>'))

lib/provider/boku.py
```diff
@@ -33,6 +33,10 @@ def create(self, request):
         cleaned = form.cleaned_data
         transaction = cleaned['param']
 
+        # Check the signature shows it was from Boku.
+        log.info('Checking signature')
```
kumar303: maybe add the transaction to this log message too

lib/provider/boku.py
```diff
@@ -33,6 +33,10 @@ def create(self, request):
...
+        check_sig(request.DATA)
```
kumar303: I don't see a test that executes this view. I think it would be nice to have a test like that, at least something with a mock of ``check_sig`` to assert that it was called.

lib/provider/boku.py
```diff
@@ -33,6 +33,10 @@ def create(self, request):
         cleaned = form.cleaned_data
         transaction = cleaned['param']
 
+        # Check the signature shows it was from Boku.
```
kumar303: typo. Maybe 'Checking...' ?

lib/provider/boku.py
```diff
@@ -33,6 +33,10 @@ def create(self, request):
...
+        log.info('Checking signature')
+        check_sig(request.DATA)
+
         # Verify this against Boku, this will raise errors if there's
         # an issue.
```
kumar303: while you're in here, I'd amend this comment to explain that it only raises an issue in certain circumstances, not for spoofed requests with valid boku transaction IDs.

lib/provider/boku.py
kumar303: if this raises a VerificationError it's probably better to log it and return a 400, no? We'd probably still want the error to go to Sentry though.

kumar303:
r- for the missing test. Other comments are minor.

lib/provider/boku.py
andymckay: Good point

lib/provider/boku.py
andymckay: Thats a good call on the test. Its a pretty critical error imho, but yeah a 400 is a good point. I'll check how it goes through the rest of the stack.

PR #319, opened 2015-02-10
Pass product name to Boku buy flow (bug 1008419)

kumar303:
@andymckay r?

### DIFF
--- a/lib/boku/client.py
+++ b/lib/boku/client.py
@@ -217,7 +217,7 @@ class BokuClient(object):
 
     def start_transaction(self, callback_url, consumer_id,
                           external_id, price, currency, service_id,
-                          forward_url, country):
+                          forward_url, country, product_name):
         """
         Begin a transaction with Boku.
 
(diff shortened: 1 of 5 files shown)


lib/boku/client.py
```diff
@@ -234,18 +234,30 @@ def start_transaction(self, callback_url, consumer_id,
...
                 'No multiplier to figure out decimal places for currency {c}'
                 .format(c=currency))
         price = int(price * constants.DECIMAL_PLACES[currency])
+
+        if len(product_name) > 20:
+            log.warning('Truncating product name for Boku. Long name: {n}. '
+                        'Transaction: {t}'.format(n=product_name,
+                                                  t=external_id))
+            product_name = product_name[0:20]
+
```
andymckay: What boku won't truncate it in the ui? bah! len is always a pain for some whacky unicode strings.

lib/boku/tests/test_client.py
```diff
@@ -206,12 +210,29 @@ def test_client_start_transaction_calls_api_with_correct_params(self):
...
+    def test_long_names_are_truncated(self):
+        product_name = 'Unicorn Dust From A Majestic Forest'
+
+        with mock.patch('lib.boku.client.BokuClient.api_call') as mock_client:
+            self.start_transaction(product_name=product_name)
+            eq_(mock_client.call_args[0][1]['desc'],
+                product_name[0:20])
+
+    def test_non_ascii_names(self):
+        product_name = u'Ivan Krsti\u0107'
```
andymckay: thanks for the unicode test

andymckay:
r+

lib/boku/client.py
kumar303: should we let them truncate it themselves? I don't trust anyone, hah. I guess the main thing is I wanted to put the warning message so maybe I'll just leave that and let Boku truncate.

PR #321, opened 2015-03-04
add in md5 sig check (bug 1130213)

andymckay:
Sorry for the noise, you've seen most of this. It got simple, then complicated, then too complicated and then I removed most of that and got back to simple. Kinda embarrasing how long that took.

Here's a test in docker with an invalid signature:

```
# curl -v -X POST http://localhost:2602/provider/boku/event/ -H "Content-Type: apprx_id": '123', "action": "billingresult", "param": "foo", "sig": "not-valid"}'
...
* HTTP 1.0, assume close after body
< HTTP/1.0 400 BAD REQUEST
< Server: WSGIServer/0.1 Python/2.7.5
< Content-Length: 34
< Vary: Accept
< ETag: "14103897df789f29e185eb62f645a0dc"
< Allow: POST, OPTIONS
< Date: Tue, 03 Mar 2015 23:44:02 GMT
< Content-Type: application/json
< 
* Closing connection #0
{"sig": ["Not a valid signature"]
```

A valid signature goes on to then verify the transaction "foo" which it then fails with amusingly enough.

### DIFF
--- a/lib/boku/client.py
+++ b/lib/boku/client.py
@@ -17,10 +17,9 @@ from django_statsd.clients import statsd
 from slumber import url_join
 
 from lib.boku import constants
-from lib.boku.errors import BokuException
+from lib.boku.errors import BokuException, SignatureError
 from lib.boku.tests import sample_xml
 from solitude.logger import getLogger
-
 
 log = getLogger('s.boku')
 
@@ -34,6 +33,9 @@ def get_boku_request_signature(secret_key, request_args):
     https://merchants.boku.com/white_label/boku
         /doclibrary/BOKU_SecurityImplementation.pdf
     """
+    if not secret_key:
+        raise ImproperlyConfigured('BOKU_SECRET_KEY not set')
+
     # Sort the pairs by their key name.
     sorted_pairs = sorted(request_args.items())
 
@@ -69,8 +71,8 @@ class BokuClient(object):
         params['sig'] = get_boku_request_signature(self.secret_key, params)
         return params
 
-    def api_call(self, path, params):
-        if 'timestamp' not in params:
+    def api_call(self, path, params, add_timestamp=True):
+        if add_timestamp and 'timestamp' not in params:
             params['timestamp'] = str(calendar.timegm(time.gmtime()))
 
         self._sign(params)
@@ -81,6 +83,12 @@ class BokuClient(object):
         )
 
         response = self._get(url)
+        if response.status_code == 204:
+            # No point in trying parse the body of 204 responses.
+            assert not response.content, (
+                'Response with status: 204, body must be empty')
+            log.warning('Not parsing empty body, status: 204')
+            return
 
         # Handle an http request failure.
         if response.status_code != 200:
@@ -89,9 +97,10 @@ class BokuClient(object):
                 status=response.status_code,
                 content=response.content,
             ))
-            raise BokuException('Request Failed: {status}'.format(
+            raise BokuException(
+                'Request Failed: {status}'.format(status=response.status_code),
                 status=response.status_code
-            ))
+            )
 
         # Handle an XML parse failure.
         try:
@@ -283,6 +292,12 @@ class BokuClient(object):
             'paid': Decimal(tree.find('paid').text),
         }
 
+    def check_sig(self, *args, **kw):
+        """
+        Not checking the signature unless you are in a proxy.
+        """
+        log.info('Not checking the Boku signature locally.')
+
 
 mocks = {
     'price': (200, sample_xml.pricing_request),
@@ -328,10 +343,25 @@ class ProxyClient(BokuClient):
         with statsd.timer('solitude.boku.api'):
             return requests.get(proxy)
 
+    def check_sig(self, data):
+        """
+        Check that a signature is valid. This check has to be done on the
+        proxy, because the solitude database does not have access to the
+        required BOKU_SECRET_KEY.
+        """
+        try:
+            self.api_call('/check_sig', data, add_timestamp=False)
+        except BokuException as e:
+            # If the signature check failed, a 400 is raised.
+            if e.status == 400:
+                raise SignatureError('sig verification failed')
+            # Re-raise any other error since we aren't sure what it is.
+            raise
+
 
 def get_client(*args):
     """
-    Use this to get the right client and communicate with Bango.
+    Use this to get the right client and communicate with Boku.
     """
     if settings.BOKU_MOCK:
         log.warning('Boku is using the mock client')
--- a/lib/boku/forms.py
+++ b/lib/boku/forms.py
@@ -1,14 +1,14 @@
 from django import forms
 
+from lib.boku import constants
+from lib.boku.client import BokuClientMixin
 from lib.boku.constants import CURRENCIES
+from lib.boku.errors import BokuException, SignatureError
+from lib.boku.utils import fix_price
+from lib.sellers.models import Seller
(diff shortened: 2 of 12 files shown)


lib/boku/client.py
```diff
@@ -69,8 +71,8 @@ def _sign(self, params):
         params['sig'] = get_boku_request_signature(self.secret_key, params)
         return params
 
-    def api_call(self, path, params):
-        if 'timestamp' not in params:
+    def api_call(self, path, params, add_timestamp=True):
+        if add_timestamp and 'timestamp' not in params:
             params['timestamp'] = str(calendar.timegm(time.gmtime()))
```
kumar303: could also be:

    if add_timestamp:
        params.setdefault('timestamp', str(calendar.timegm(time.gmtime())))

lib/boku/client.py
```diff
@@ -81,6 +83,9 @@ def api_call(self, path, params):
         )
 
         response = self._get(url)
+        if response.status_code == 204:
+            # No point in trying parse the body of 204 responses.
```
kumar303: it wouldn't be the first time if a vendor sent the wrong http code. Could you add some logging here so we know why it was ignored?

lib/boku/client.py
```diff
@@ -283,6 +289,11 @@ def check_transaction(self, transaction_id):
             'paid': Decimal(tree.find('paid').text),
         }
 
+    def check_sig(self, *args, **kw):
+        """
+        Not checking the signature unless you are in a proxy.
+        """
```
kumar303: a log warning here like "not checking signature" might be nice

lib/proxy/views.py
```diff
@@ -238,6 +238,26 @@ def sign(self, request):
...
+def check_sig(request):
+    """
+    Override the check_sig call, Boku doesn't actually implement this,
+    the proxy does because it has access to that data. Rather than
+    send the data on, or try overriding the client, just grab this request,
+    parse it and send back a 204 or 400.
+    """
+    data = request.GET.copy()
+    sig = data.pop('sig')[0]
+    print data
```
kumar303: :fried_shrimp:

lib/proxy/views.py
```diff
@@ -238,6 +238,26 @@ def sign(self, request):
...
+    value = get_boku_request_signature(settings.BOKU_SECRET_KEY, data)
+    print value, sig
```
kumar303: :fries:

lib/boku/client.py
```diff
@@ -328,10 +339,25 @@ def _get(self, url):
...
             return requests.get(proxy)
 
+    def check_sig(self, data):
+        """
+        Check that a signature is valid. This check has to be done on the
+        proxy, because the solitude database does not have access to the
+        required BOKU_SECRET_KEY.
+        """
+        try:
+            self.api_call('/check_sig', data, add_timestamp=False)
```
kumar303: why does this need to make a fake API call? Can't you just move [this code](https://github.com/andymckay/solitude/blob/1130213/lib/proxy/views.py#L242-L255) into this function?

lib/boku/forms.py
```diff
@@ -55,6 +54,19 @@ def clean(self):
                                                 cleaned_data.get('currency')))
         return cleaned_data
 
+    def clean_sig(self):
+        # Check that the signature was from Boku. This will raise an error
+        # if the signing is incorrect, but not check the transaction_id is
+        # valid.
```
kumar303: This reads awkwardly to me. Maybe: "This will raise an error if the signature is incorrect but it does not validate the transaction_id."

lib/provider/tests/test_boku.py
```diff
@@ -38,7 +38,7 @@ def test_good(self):
     @patch('lib.boku.client.mocks', {'verify-trx-id': (500, '')})
     def test_verify_fails(self):
         self.add_seller_boku()
-        eq_(self.client.post(self.url, data=self.sample()).status_code, 200)
+        self.client.post(self.url, data=self.sample()).status_code
```
kumar303: Why do you need to access ``.status_code`` ? Does that force-parse the response or something? If so, is there a more explicit way to do that?

lib/proxy/views.py
```diff
@@ -238,6 +238,26 @@ def sign(self, request):
...
+    """
+    Override the check_sig call, Boku doesn't actually implement this,
+    the proxy does because it has access to that data. Rather than
+    send the data on, or try overriding the client, just grab this request,
+    parse it and send back a 204 or 400.
+    """
+    data = request.GET.copy()
+    sig = data.pop('sig')[0]
+    print data
+    value = get_boku_request_signature(settings.BOKU_SECRET_KEY, data)
```
kumar303: if you renamed ``value`` to ``internal_sig`` and renamed ``sig`` to ``external_sig`` then maybe it would read better

lib/proxy/views.py
```diff
@@ -238,6 +238,26 @@ def sign(self, request):
...
 
+def check_sig(request):
+    """
+    Override the check_sig call, Boku doesn't actually implement this,
+    the proxy does because it has access to that data. Rather than
+    send the data on, or try overriding the client, just grab this request,
+    parse it and send back a 204 or 400.
+    """
+    data = request.GET.copy()
+    sig = data.pop('sig')[0]
```
kumar303: won't this raise a KeyError when no ``sig``? Or are you relying on the form to protect against this?

lib/proxy/views.py
```diff
@@ -238,6 +238,26 @@ def sign(self, request):
...
+    print data
+    value = get_boku_request_signature(settings.BOKU_SECRET_KEY, data)
+    print value, sig
+    result = value == sig
```
kumar303: maybe rename ``result`` to ``is_valid``

kumar303:
r- unless you can convince me that we really need a fake API call for this

lib/proxy/views.py
andymckay: poop

lib/proxy/views.py
andymckay: sure

lib/proxy/views.py
andymckay: yeah, it won't get to the proxy because the EventForm kills it

lib/provider/tests/test_boku.py
andymckay: Ah, we don't - I will remove that. This was unrelated to the md5 stuff, the code actually checks a `BokuValidation` is raised, the `eq_(....status_code, 200)` is a distraction in the test.

lib/boku/tests/test_client.py
```diff
@@ -322,7 +342,7 @@ def setUp(self):
 
     def test_good(self):
         # A copy of
-        # test_client_check_transaction_returns_check_transaction_json
+        # test_client_check_transaction_retutestrns_check_transaction_json
```
andymckay: hey wat?

lib/boku/client.py
andymckay: It's not a Fake API call, its a call to the Proxy. The ProxyClient is run by the database instance to interact with the Boku proxy.

lib/boku/client.py
andymckay: sure

lib/boku/client.py
andymckay: ok

andymckay:
Confused by the reason for r- as per comments.

kumar303:
whoops, the proxy melted my brain. I get it now. r+!

PR #322, opened 2015-03-07
allow transactions to be created and errored before provider is known (bug 1140615)

andymckay:
* allows a transaction to have no provider, so we can report transaction failures when there is no error
* adds in two new states
* makes ERRORED and end state, that you cannot update from

### DIFF
--- a/lib/transactions/forms.py
+++ b/lib/transactions/forms.py
@@ -15,9 +15,6 @@ log = getLogger('s.transaction')
 
 
 def check_status(old, new):
-    if old['status'] == constants.STATUS_ERRORED:
-        raise forms.ValidationError('Transaction errored')
-
     if ((old['created'] + timedelta(seconds=settings.TRANSACTION_LOCKDOWN)) <
             datetime.now()):
         raise forms.ValidationError('Transaction locked down')
@@ -25,7 +22,7 @@ def check_status(old, new):
     elif old['status'] == constants.STATUS_PENDING:
         return
 
-    elif old['status'] in [constants.STATUS_FAILED, constants.STATUS_CANCELLED,
+    elif old['status'] in [constants.STATUS_FAILED, constants.STATUS_ERRORED,
                            constants.STATUS_CANCELLED]:
         msg = 'Cannot change state: {0}'.format(old['status'])
         log.error(msg)
@@ -53,8 +50,11 @@ class UpdateForm(ParanoidForm):
     notes = forms.CharField(required=False)
     status = forms.ChoiceField(choices=[(v, v) for v in STATUSES.values()],
                                required=False)
+    status_reason = forms.CharField(required=False)
     uid_pay = forms.CharField(required=False)
     pay_url = forms.URLField(required=False)
+    provider = forms.ChoiceField(choices=constants.PROVIDERS_CHOICES,
+                                 required=False)
 
     def __init__(self, *args, **kw):
         self.request = kw.pop('request')  # Storing request for CEF logs.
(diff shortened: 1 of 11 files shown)


andymckay:
@kumar303 I think this first step is good to go now.

lib/transactions/forms.py
```diff
@@ -51,7 +62,12 @@ def __init__(self, *args, **kw):
         super(UpdateForm, self).__init__(*args, **kw)
 
     def clean(self):
-        keys = set(self.data.keys()).difference(set(self.fields.keys()))
+        fields = self.fields.keys()
+        # If the provider is already set, don't allow it to be changed.
+        if self.old.get('provider'):
```
kumar303: Can this maybe raise a validation error? I don't like data that silently disappears. Or at least log a warning?

lib/transactions/tests/test_api.py
```diff
@@ -51,8 +60,8 @@ def test_post_uuid(self):
                 seller_id=self.sellers.seller.pk),
             'buyer': '/generic/buyer/{buyer_id}/'.format(
                 buyer_id=self.buyer.pk),
+            'status_reason': 'ALL_COOL'
```
kumar303: trailing commas are nice for copy/paste

lib/transactions/tests/test_api.py
```diff
@@ -126,3 +152,21 @@ def test_relations(self):
         eq_(data['objects'][0]['relations'], [])
         eq_(data['objects'][0]['related'],
             '/generic/transaction/%s/' % self.trans.pk)
+
+    def test_create_minimal(self):
+        res = self.client.post(self.list_url, data={})
+        eq_(res.status_code, 201)
```
kumar303: maybe check the default status here?

lib/transactions/tests/test_forms.py
```diff
@@ -35,22 +38,46 @@ def test_failed(self):
...
+                      'status': constants.STATUS_STARTED},
+                     {'status': constants.STATUS_ERRORED})
+
+    def test_completed_not_ok(self):
+        with self.assertRaises(ValidationError):
+            check_status({'created': datetime.now(),
+                          'status': constants.STATUS_STARTED},
+                         {'status': constants.STATUS_COMPLETED})
+
+        with self.assertRaises(ValidationError):
```
kumar303: could this be a separate test?

migrations/52-add-error-reason.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE transaction ADD COLUMN status_reason varchar(255);
```
kumar303: this should be NULL, right?

kumar303:
r+wc

lib/transactions/forms.py
andymckay: It doesn't delete the data, if deletes the field off the form. So if you set provider and then try and patch provider it throws a ValidationError on the next lines (https://github.com/mozilla/solitude/pull/322/files#diff-ecc0a1a041d0a2d3c4e9555c196ab6b8R73). The deleting of the field off the form is an internal thing, but this is a quick way to ensure that from a calling API point of view, you'll get a 400 on the second attempt. I've added a test to demonstrate this.

lib/transactions/tests/test_api.py
andymckay: JS has ruined me.

migrations/52-add-error-reason.sql
andymckay: varchar columns are nullable, so mysql makes them null by default, I can add to be explicit though :)

PR #327, opened 2015-03-20
remove paypal (bug 1141700)

andymckay:
Note: database tables still exist, these will be removed on a later commit to prevent any deploy issues.

* [x] see if these were doing something useful https://github.com/mozilla/solitude/pull/327/files#diff-d7481aba4d9a61cd39f740242a04d013


lib/bango/tests/test_resources.py
```diff
@@ -603,7 +602,7 @@ def test_with_product_icon(self):
     def test_create_trans_if_not_existing(self):
         data = self.good()
         data['transaction_uuid'] = '<some-new-trans-uuid>'
-        self.transaction.provider = constants.PROVIDER_PAYPAL
+        self.transaction.provider = PROVIDER_BOKU
```
kumar303: maybe use Bango for generic tests? Boku may not get much real world usage :/

kumar303: oh nevermind, I see that the provider needs to not be Bango for these tests

lib/transactions/constants.py
```diff
@@ -41,13 +41,11 @@
 TYPE_REVERSALS = (TYPE_REVERSAL, TYPE_REVERSAL_MANUAL)
 TYPE_REFUNDS_REVERSALS = TYPE_REFUNDS + TYPE_REVERSALS
 
-PROVIDER_PAYPAL = 0
```
kumar303: I think Webpay will need some cleanup too: https://github.com/mozilla/webpay/blob/master/lib/solitude/constants.py#L23

lib/transactions/utils.py
```diff
@@ -1,68 +0,0 @@
...
-                constants.STATUS_PENDING,
-                constants.STATUS_CHECKED
-            )
-        )
-    except Transaction.DoesNotExist:
-        return False
-
-    record.status = constants.STATUS_COMPLETED
-    record.save()
-    return True
```
kumar303: none of this is used anymore? It doesn't look paypal specific at a glance

solitude/base.py
```diff
@@ -186,7 +186,6 @@ def allowed_verbs(self, url, allowed):
         Otherwise the test will fail.
         """
         verbs = ['get', 'post', 'put', 'delete', 'patch']
-        # TODO(andym): get patch in here.
```
kumar303: :+1:

solitude/settings/test.py
```diff
@@ -9,11 +9,7 @@
 filename = os.path.join(os.path.dirname(__file__), 'sample.key')
 
 AES_KEYS = {
-    'buyerpaypal:key': filename,
     'buyeremail:key': filename,
-    'sellerpaypal:id': filename,
-    'sellerpaypal:token': filename,
-    'sellerpaypal:secret': filename,
```
kumar303: I wonder if we should ask ops to delete all these files

lib/transactions/utils.py
andymckay: It doesn't, but the only place that calls it is the Paypal IPN code. Bango and Boku have their own event notice code implemented. Removing this didn't cause any errors in those code paths.

lib/transactions/constants.py
andymckay: Good call, will do.

kumar303:
r+wc

hooray for deleting unnecessary code and tests

lib/bango/tests/test_resources.py
andymckay: I also wanted to get out of the trap of using Bango for everything then have it all fail when we add in something not Bango (that happened a bit with Boku).

solitude/settings/test.py
andymckay: Won't hurt, will do.

PR #328, opened 2015-03-23
use unmodified data for sig check (bug 1141640-two)

andymckay:


kumar303:
Can you add a test that uses a field with a dash in the name so that the signature check fails unless you use the unmodified data?

andymckay:
The test I altered does that now, it's what the test was meant to be doing. But `is_called_with` doesn't exist and passes because it calls the mock. Fixing the test shows the error.

kumar303:
ok, if the test fails when you swap in ``self.data`` then I'm happy. r+

kumar303:
> But is_called_with doesn't exist and passes because it calls the mock.

Sigh. I want to stop that from ever happening somehow.

PR #329, opened 2015-03-23
move implicit bango transaction creation (bug 1036579)

andymckay:
* [x] will need to wait for matching pull request in webpay 

See also: https://github.com/mozilla/webpay/pull/615


kumar303:
I like all the red. Most of my related comments are on https://github.com/mozilla/webpay/pull/615 so I am r+ here.

PR #332, opened 2015-03-27
remove create signal as well (bug 1036579)

andymckay:
Follow up to removing the implicit transaction, we no longer need the create signal and some of the data passed through to the form.


kumar303:
r+

PR #336, opened 2015-04-01
PyJWT 0.1.4.2 no longer exists on pypi

andymckay:


kumar303:
This version is insecure so it can't be used. Huh. I don't think it's used in solitude anyway. I created https://github.com/mozilla/solitude/pull/337

PR #339, opened 2015-04-03
allow carrier and region to be set on a transaction (bug 926575)

andymckay:


kumar303:
Could you maybe add those fields to an existing test?

r+wc

andymckay:
I couldn't think of one that made sense really.

PR #340, opened 2015-04-10
use peep and delete some cruft (fixes #21)

andymckay:


Dockerfile
```diff
@@ -11,7 +11,9 @@ ENV IS_DOCKER 1
 RUN mkdir -p /pip/{cache,build}
 ADD requirements /pip/requirements
 WORKDIR /pip
-RUN pip install -b /pip/build --download-cache /pip/cache --no-deps -r /pip/requirements/dev.txt --find-links https://pyrepo.addons.mozilla.org/
+# Download this securely from pyprepo first.
+RUN pip install --no-deps --find-links https://pyrepo.addons.mozilla.org/ peep
```
kumar303: could you put ``peep>=X.X`` so that if it breaks something we at least know the last known working version?

requirements/prod.txt
```diff
@@ -2,50 +2,128 @@
 # needed to run the site, it belongs in requirements/prod.txt. If it's a
 # package for developers (testing, docs, etc.), it goes in this file.
 
+# sha256: 7sV9MhlQHsbmhWRoJvLrjndoepP-N0p-aZRJBSDbCT4
```
kumar303: hooray for hash checks!

kumar303:
Thanks for deleting all those old deploy scripts. I suppose we could do the same for webpay.

Dockerfile
```diff
@@ -11,7 +11,9 @@ ENV IS_DOCKER 1
 RUN mkdir -p /pip/{cache,build}
 ADD requirements /pip/requirements
 WORKDIR /pip
-RUN pip install -b /pip/build --download-cache /pip/cache --no-deps -r /pip/requirements/dev.txt --find-links https://pyrepo.addons.mozilla.org/
+# Download this securely from pyprepo first.
+RUN pip install --no-deps --find-links https://pyrepo.addons.mozilla.org/ peep==2.2
```
kumar303: peep 2.4 is available but I don't think you need to pin this to a version. Unlike other libs, this is one you want to keep up to date. I doubt the CLI interface will break but if that happens you could go back and pin it.

kumar303:
r+wc

Dockerfile
andymckay: That sounds reasonable, pinning to a version is something I'm used to :)

PR #345, opened 2015-04-15
fix boku delete test (fixes #344)

### LINKED ISSUE #344: Fix boku test
Surprised that this passes, but I think its because the pk's are different when entire test suite is run?

`test_delete_removes_seller_product_boku` creates a `SellerProductBoku` object, then deletes it. It then tests that a `SellerProduct` object of the same pk as the created one does not exist. It should be checking that a `SellerProductBoku` does not exist and this test will pass if for some case the pk's don't line up.

andymckay:


kumar303:
Ha. Oops.

r+

PR #354, opened 2015-04-20
stand up braintree libs and settings (fixes #350)

### LINKED ISSUE #350: Add in braintree libs and settings
Add in initial braintree libs and settings to solitude.

Supports https://github.com/mozilla/payments/issues/16

andymckay:
```
bash-4.1# python manage.py  shell
[snip]
In [1]: from lib.brains.client import get_client

In [2]: from django.http import Http404

In [3]: client = get_client()

In [4]: try:
   ...:     customer = client.Customer.find('123')
   ...: except Http404:
   ...:     print 'not found'
   ...:     
not found
```

Just some initial config.


lib/brains/client.py
```diff
@@ -0,0 +1,95 @@
...
+            if len(response_body.strip()) == 0:
+                return {}
+            else:
+                return XmlUtil.dict_from_xml(response_body)
+
+
+class OurConfiguration(Configuration):
+
+    def http(self):
+        return Http(self)
```
kumar303: Did you try just passing in ``http_strategy`` to ``Configuration()``? It looks like you won't need a custom class that way: https://github.com/braintree/braintree_python/blob/master/braintree/configuration.py#L70

lib/brains/client.py
```diff
@@ -0,0 +1,95 @@
...
+            else:
+                raise e
+
+        if Http.is_error_status(status):
+            Http.raise_exception_from_status(status)
+        else:
+            if len(response_body.strip()) == 0:
+                return {}
+            else:
+                return XmlUtil.dict_from_xml(response_body)
```
kumar303: This overrides way more than we need to which will be cumbersome to maintain. I know you're dealing with very java-esque Python but it looks like ``http_strategy`` has all the hooks you need. What about something like this:

````
class Http(braintree.util.http.Http):

    def http_do(self, *args, *kw):
        with statsd.timer('solitude.braintree.api'):
            return super(Http, self).http_do(*args, **kw)

    def handle_exception(self, exc):
        if exc.status == 404:
            raise Http404
        raise

Configuration(http_strategy=Http, wrap_http_exceptions=true, ...)
````

File: lib/brains/client.py [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+
+    @staticmethod
+    def raise_exception_from_status(status, message=None):
+        """
+        It would be nice if a 404 from Braintree became a 404 from Django
+        as opposed to a 500 error.
+
+        This method allows us to override the exception response.
+        """
+        statsd.incr('solitude.bango.response.{0}'.format(status))
```
kumar303: not bango ;)

kumar303: also, these are only exceptions which I don't think is what you intended for statsd

kumar303:
r- on all the overrides. If my alternate idea doesn't work then let me know; I have other ideas.

kumar303:
r+

Thanks, that looks a lot simpler and should be easier to maintain.

PR #357, opened 2015-04-20
start of create and check braintree config (fixes #352)

### LINKED ISSUE #352: Add in braintree configuration
Add commands to create braintree subscription products from the command line, based on a JSON or other configuration file.

If run multiple times, should not fail, but instead just verify that things are ok.

https://github.com/mozilla/payments/issues/16

andymckay:


andymckay:
fixes #352

solitude/settings/products.py
```diff
@@ -0,0 +1,20 @@
+# A definiton of Products for Payments for Firefox Accounts. Marketplace
+# defines these dynamically. However for the moment we will define these from
+# configuration files.
+from decimal import Decimal
+
+products = {
```
kumar303: why can't this just be ``BRAINTREE_PRODUCTS`` or something in the normal settings? That way you won't need all that import code.

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+        raise CommandError('Cannot find "products" in file {0}'.format(path))
+    return products
+
+
+def get_or_create_seller(uuid):
+    """
+    Create a seller in solitude for the product.
+    """
+    # We'll suffix with the BRAINTREE_PUBLIC_ID so that the seller will
+    # change if you change ids, or servers.
```
kumar303: A UUID is unique so how does this comment apply?

kumar303: are you trying to prevent accidental lookups somehow?

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+def product_exists(plans, external_id, amount):
+    """
+    Check that the product exists in Braintree.
+    """
+    if external_id not in plans:
+        log.warning(
+            'The plan: {0} does not exist in Braintree. Plans cannot be '
+            'created from the API and must be created from the Braintree '
+            'console.'
+            .format(external_id))
```
kumar303: I think it would be fine if you just raised a CommandError with this message

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+            seller = get_or_create_seller(item['seller'])
+            # Iterate through each product checking they exist.
+            for product in item['products']:
+                external_id = item['seller'] + '-' + product['name']
+                get_or_create_seller_product(
+                    external_id=external_id,
+                    public_id=product['name'],
+                    seller=seller
+                )
+                product_exists(get_plans(), external_id, product['amount'])
```
kumar303: this would read better to me as ``validate_product(...)``

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+            '--path',
+            dest='path',
+            default='solitude.settings.products',
+            help=('Dotted path to the settings file, '
+                  'eg: solitude.settings.products')
+        ),
+    )
+
+    def handle(self, *args, **options):
+        settings = import_file(options['path'])
```
kumar303: this ``settings`` var overrides the django settings

lib/brains/management/commands/braintree_config.py
kumar303: oh I see, it's not really a UUID, it's a name

andymckay: Ok so then its not a uuid and adding in the braintree_public_key should make it a uuid. I'm trying to ensure that if you flip your local solitude accounts around in development all the data doesn't get attached to the wrong rows. For example: do some work in account X, flip to account Y and you wouldn't want to start attaching sales to the seller created in account X.

solitude/settings/products.py
```diff
@@ -0,0 +1,20 @@
...
+    'concrete': {
+        'seller': 'mozilla-concrete',
+        'products': [
+            {
+                'name': 'brick',
+                'amount': Decimal('10')
```
kumar303: aren't these going to need more definition like recurring subscription fee and period of validity? I guess we could amend it later.

lib/brains/management/commands/braintree_config.py
kumar303: It looks like you don't need to call ``get_plans()`` on every loop

solitude/settings/products.py
andymckay: Wasn't sure how big or complicated this was going to get. I wanted something small and self contained that people could read and update without having to plow through settings. But then again it could just be a seperate file that we import.

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+        for name, item in settings.items():
+            log.info('Configuring: {0}'.format(name))
+            seller = get_or_create_seller(item['seller'])
+            # Iterate through each product checking they exist.
+            for product in item['products']:
+                external_id = item['seller'] + '-' + product['name']
+                get_or_create_seller_product(
+                    external_id=external_id,
```
kumar303: honestly, we probably won't ever use this field for braintree but I guess it won't hurt to populate it.

lib/brains/management/commands/braintree_config.py
```diff
@@ -0,0 +1,108 @@
...
+                    public_id=product['name'],
```
kumar303: this column current has a unique flag on it meaning unique across the entire table. This might need to be seller + name too.

solitude/settings/products.py
kumar303: I think I see. This isn't a configuration used to create the objects. So amount is just to do some extra validation?

lib/brains/tests/test_management.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @raises(CommandError)
+    def test_bad(self):
+        config.import_file('foo.bar')
+
+    def test_good(self):
+        ok_(config.import_file('solitude.settings.products'))
+
+    def test_created(self):
+        seller = config.get_or_create_seller('uuid:fxa')
```
kumar303: not sure what ``fxa`` is about in this example

kumar303: could this be more realistic like ``config.get_or_create_seller('mozilla-concrete')`` ?

lib/brains/tests/test_management.py
```diff
@@ -0,0 +1,48 @@
...
+        same_seller = config.get_or_create_seller('uuid:fxa')
+        eq_(seller, same_seller)
+
+    def test_created_product(self):
+        seller = config.get_or_create_seller('uuid:fxa')
+        seller_product = config.get_or_create_seller_product(
+            external_id='some:product:uuid:fxa',
+            public_id='some:product',
+            seller=seller)
+        eq_(seller_product, seller.product.get())
```
kumar303: what is this testing? Can the function name be more specific? I'm not sure what this assertion is proving.

kumar303: oh I read assertion wrong. I think I see what it's doing.

solitude/settings/products.py
kumar303: it seems a little weird to me that if you want to change the price of a product you'd have to do so in both braintree and the config file.

solitude/settings/products.py
kumar303: I think it's complicated as a standalone file. How do we define dev vs prod products? We're going to have to reinvent a lot of Django.

lib/brains/management/commands/braintree_config.py
kumar303: Sounds right. I see how it works now.

kumar303:
Cool to have a command for this.

r+wc

solitude/settings/products.py
andymckay: Yeah i was disappointed to find that we can't create plans in the API, that was the goal of this command, it could now be useless.

But I think a general, "build out the required db values in dev and then check your config in braintree is ok" before we start to do anything might be good. We had a similar script for Bango that proved useful with ops multiple times.

solitude/settings/products.py
andymckay: Pulling it from settings is fine, I'll change that.

lib/brains/management/commands/braintree_config.py
andymckay: i don't, thanks

lib/brains/tests/test_management.py
andymckay: sure

PR #358, opened 2015-04-21
add in mock client (fixes #355)

### LINKED ISSUE #355: Set up a braintree mock
So that tests can be run without hitting Braintree servers.

andymckay:


kumar303:
test failures look like pep8 lint

lib/brains/client.py
```diff
@@ -0,0 +1,142 @@
...
+            else:
+                return XmlUtil.dict_from_xml(response_body)
+
+# To add in more data into the samples directory to populate mocks:
+# * set settings.DUMP_REQUESTS to Tru
+# * run the Django shell
+# * run commands to access braintree
+# * copy data from the logs into samples files
+mocks = {
+    'GET /merchants//customers/does-not-exist': (404, ''),
```
kumar303: do all these paths need the double slashes?

lib/brains/tests/samples/customer.py
```diff
@@ -0,0 +1,257 @@
...
+  <phone nil="true"/>
+  <fax nil="true"/>
+  <website nil="true"/>
+  <created-at type="datetime">2015-04-03T15:20:04Z</created-at>
+  <updated-at type="datetime">2015-04-03T15:20:04Z</updated-at>
+  <custom-fields/>
+  <credit-cards type="array"/>
+  <addresses type="array"/>
+</customer>
+"""
```
kumar303: This python module is going to get pretty wild. I suggest making them into xml files like ``samples/customers/minimal.xml``, ``samples/customers/full.xml``, etc, and just having ``http_do()`` load the XML files.

lib/brains/client.py
```diff
@@ -0,0 +1,142 @@
...
+        'production': braintree.Environment.Production,
+    }
+
+    braintree.Configuration.instantiate = instantiate
+    braintree.Configuration.configure(
+        environments[settings.BRAINTREE_ENVIRONMENT],
+        settings.BRAINTREE_MERCHANT_ID,
+        settings.BRAINTREE_PUBLIC_KEY,
+        settings.BRAINTREE_PRIVATE_KEY
+    )
```
kumar303: Can't you just do this:

````
if settings.BRAINTREE_MOCK:
    braintree.Configuration.configure(http_strategy=MockHttp, ...)
````

File: lib/brains/client.py [hunk 4]
```diff
@@ -0,0 +1,142 @@
...
+
+def get_client():
+    """
+    Use this to get the right client and communicate with Braintree.
+    """
+    braintree._http_configuration_class = NormalConfiguration
+
+    if settings.BRAINTREE_MOCK:
+        log.warning('Braintree is using the mock client')
+        braintree._http_configuration_class = MockConfiguration
```
kumar303: this seems brittle and unnecessary. See below for the ``http_strategy`` idea.

kumar303:
r- so far on accessing and overriding private braintree functions. Let me know if the ``http_strategy`` idea doesn't work.

kumar303:
My opposition to this is it makes the source of the mock data too magical. There is no way to look at test code that does ``braintree.Customers.get('minimal')`` and be able to easily figure out where minimal is defined or how to change it.

Mocking is painful either way but I think patching would be more clear. We could easily make a base class that patched what was needed and write tests like this:

````
class TestCustomerGetting(Base, WithMinimalCustomer, WithFullCustomer):

    def test(self):
        ...
````

That way you could at least trace the base classes to see where/how the API response was patched for this specific test.

r+wc

PR #360, opened 2015-04-22
don't change uuid if not given (bug 1157054)

andymckay:


lib/transactions/serializers.py
```diff
@@ -46,6 +46,6 @@ def resource_uri(self, pk):
 
     def validate_uuid(self, attrs, source):
         # Provide a default uuid.
-        if not attrs.get('uuid'):
+        if not attrs.get('uuid') and not self.object:
```
kumar303: Is ``self.object`` always defined? Or do you need ``getattr(self, 'object')`` ?

kumar303:
r+wc

lib/transactions/serializers.py
andymckay: its always defined, its None if there's no object

PR #363, opened 2015-04-23
add in packageId properly

andymckay:


kumar303:
This sounds like a really good time to add a mock assertion to this test for packageId https://github.com/mozilla/solitude/blob/master/lib/bango/tests/test_resources.py#L590

r+wc

Issue #369, opened 2015-04-23
Add in ClientToken.generate


andymckay-limited-access:
So that the front end can get a token.

andymckay-limited-access:
Done in https://github.com/mozilla/solitude/pull/371

PR #371, opened 2015-04-24
add in ClientToken.generate. #369.

andymckay:


docs/topics/braintree.rst
```diff
@@ -3,4 +3,23 @@
...
 
-More to come.
+Generate a token
+----------------
+
+Calls braintree `ClientToken.generate <https://developers.braintreepayments.com/javascript+python/reference/request/client-token/generate>`_:
+
+.. http:post:: /braintree/token/generate/
+
+    **Request**
```
kumar303: I feel like we're going to need to pass in ``customer_id`` but we could always add that later. I'm thinking we'll want it when dealing with saved customer cards.

lib/brains/client.py
```diff
@@ -10,7 +10,10 @@
 
 log = getLogger('s.brains')
 
+# Note: between the double slash is where the BRAINTREE_MERCHANT_ID goes. In
+# tests this will be blank, hence the double slash.
```
kumar303: ah, thanks for the note

lib/brains/views.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from lib.brains.client import get_client
+
+
+@api_view(['POST'])
```
kumar303: all the authentication here is included by default?

docs/topics/braintree.rst
andymckay: Yup can add. Reading the docs it felt like it was only used for the drop-in UI which we don't use.

kumar303:
Did you forget to add the test file?

docs/topics/braintree.rst
kumar303: Right. I'm pretty sure we'll want to do saved cards though -- we may have to switch to the drop-in for that. I suppose it should be a buyer uuid.

lib/brains/views.py
andymckay: yup, but thanks for reminding me: https://github.com/mozilla/solitude/issues/372

docs/topics/braintree.rst
andymckay: You get that information by querying the Customer API in Braintree. For our purposes we'll be storing it in solitude along with the buyer, so it will come back when you query the buyer.

lib/brains/tests/test_token.py
```diff
@@ -0,0 +1,16 @@
...
+from nose.tools import eq_
+from django.core.urlresolvers import reverse
+
+
+class TestToken(APITest):
+
+    def setUp(self):
+        self.url = reverse('braintree:token.generate')
+
+    def test_token(self):
```
kumar303: I know it's a bit paranoid but it might be nice to add a test from an anonymous client that asserts a 403

kumar303:
r+wc

lib/brains/tests/test_token.py
andymckay: I've got those around auth already, so I feel pretty comfortable on that, I know every end point is protected.

PR #373, opened 2015-04-24
fix .json on test results (fixes #370)

### LINKED ISSUE #370: Fix json parser in tests
Doesn't quite work right, copy over zambonis. Then delete all that json.loads(...) in tests.

andymckay:


kumar303:
Oh right. We should really submit a patch to Django for this.

r+

andymckay:
Fixes https://github.com/mozilla/solitude/issues/370

PR #375, opened 2015-04-27
copy in the source

andymckay:


kumar303:
I guess it would be useful for deployment (even though we may not be doing that yet for security reasons).

r+

PR #376, opened 2015-04-27
move mocks into the tests

andymckay:


lib/brains/client.py
```diff
@@ -10,14 +10,17 @@
...
 
-# Note: between the double slash is where the BRAINTREE_MERCHANT_ID goes. In
-# tests this will be blank, hence the double slash.
-mocks = {
-    'POST /merchants//client_token': (200, get_sample('token')),
-    'GET /merchants//customers/does-not-exist': (404, ''),
-    'GET /merchants//customers/minimal': (200, get_sample('customer-minimal')),
-    'GET /merchants//customers/full': (200, get_sample('customer-full')),
-}
+mocks = {}
```
kumar303: maybe a comment above this to explain that a test is responsible for populating it

lib/brains/tests/test_client.py
```diff
@@ -22,7 +22,9 @@ def test_normal(self):
 
     @raises(NotFoundError)
     def test_not_found(self):
+        set_mocks(['GET', 'customers/does-not-exist', 404, ''])
```
kumar303: this is awesome. It makes it way more clear how a test gets its data. You could even roll them up in shared methods like ``setup_non_existant_customer()`` and I think it will still be clear.

lib/brains/tests/test_client.py
```diff
@@ -22,7 +22,9 @@ def test_normal(self):
```
kumar303: I think the mock data should be torn down after each test. This way your tests wouldn't have to worry about coming up with a unique ID and different tests could mock different responses for the same endpoint (500, 404, etc)

kumar303:
r+wc

lib/brains/tests/test_client.py
andymckay: The mocks get cleared on each set_mocks call, so you can do it multiple times in a test and do the different responses.

Clearing mock data on tear down is a good policy though.

Issue #374, opened 2015-04-24
Add in customer


andymckay-limited-access:
Add in a buyer and customer into solitude and braintree

PR #377, opened 2015-04-27
add in a create braintree and customer end point (fixes mozilla/solitude...

andymckay:
...#374)


andymckay:
adding in braintree - buyer model is next and will use the view here

docs/topics/generic.rst
```diff
@@ -5,6 +5,8 @@ Generic
 
 The generic API can be used for buyers and sellers.
 
+.. _buyer-label:
```
kumar303: Label is redundant. You could just say ``_buyer:``

lib/brains/client.py
```diff
@@ -37,10 +37,16 @@ class HttpMock(Http):
...
     def http_do(self, http_verb, full_path, headers, body):
         try:
-            return mocks['{0} {1}'.format(http_verb, full_path)]
+            next_mock = mocks.pop()
+            if next_mock[:2] != (http_verb, full_path):
+                # The mock was called in an order you didn't expect.
+                raise MockError(
+                    'Mock called in wrong order, expecting: {0} {1}'
+                    .format(http_verb, full_path))
+            return next_mock[2:]
```
kumar303: hmm. Is this really better than working with mock patches? You can iterate calls with ``patched_customers.find.side_effect = [..., ...]``

lib/brains/tests/samples/customer-error.xml
```diff
@@ -0,0 +1,21 @@
...
+      </errors>
+    </customer>
+  </errors>
+  <params>
+    <customer>
+      <id>sdfsdf:sdf:</id>
+    </customer>
+  </params>
+  <message>Customer ID is invalid (use only letters, numbers, '-', and '_').</message>
+</api-error-response>
```
kumar303: This feels like a really heavy fixture when a mock patch would probably only require something like:
````
result = mock.Mock()
result.is_success = False
result.message = 'Customer ID is invalid...'
patched_customer.create.return_value = result
````

File: docs/topics/braintree.rst [hunk 4]
```diff
@@ -10,16 +10,47 @@ Calls braintree `ClientToken.generate <https://developers.braintreepayments.com/
 
 .. http:post:: /braintree/token/generate/
 
-    **Request**
+    :>json string token: the token returned by Braintree.
```
kumar303: I think this is a regular urlencoded form param not a JSON object, right?

kumar303: oh nevermind, git diff was misleading me

docs/topics/braintree.rst
```diff
@@ -10,16 +10,47 @@ Calls braintree `ClientToken.generate <https://developers.braintreepayments.com/
...
+
+Creates a customer in Braintree and the corresponding buyer in solitude. If the
+buyer exists in solitude already, it is not created. If the customer exists in
+Braintree already, it is not created.
 
-    No parameters.
+.. http:post:: /braintree/customer/
 
-    **Response**
+    :<json string uuid: the uuid of the buyer in solitude.
```
kumar303: Could you prefix this with ``**Request**`` so it's more clear? And below, I think ``**Response**`` instead of 'Returns...' would be more clear as well.

kumar303: also, I think ``uuid`` is a urlencoded form param not a JSON object

andymckay:
fixes #374

lib/brains/tests/samples/customer-error.xml
kumar303: Also, if this XML changes but the library interface stays the same, the person updating this code will be sad

solitude/exceptions.py
```diff
@@ -2,9 +2,18 @@
...
 from lib.bango.errors import BangoImmediateError
+from lib.brains.errors import BraintreeResultError
 
 
 def custom_exception_handler(exc):
+    if isinstance(exc, BraintreeResultError):
+        res = {
+            '__all__': [exc.result.message],
+            '__braintree__': 'error',
+            '__type__': 'braintree'
```
kumar303: why do these need to be double underscored? Can you just say ``'error_type': ...`` ?

kumar303:
The mock stuff is starting to look more and more hairy.

lots of comments in here but r+wc

solitude/exceptions.py
andymckay: That's what we've got in for some errors right now, but our errors are really inconsistent, so I filed https://github.com/mozilla/solitude/issues/349 to go through, unfortunately things in webpay and zamboni depend on formats, so it will be a bit of a hairy change.

lib/brains/client.py
andymckay: Well i don't have to do it every test, it just happens all the time now, so for the moment I'm happy.

docs/topics/braintree.rst
andymckay: screenshot 2015-04-27 15 56 16
Sphinx prepends Request or Response if it see < or > , so it seemed redundant.

lib/brains/tests/samples/customer-error.xml
andymckay: Thats a valid point.

docs/topics/braintree.rst
kumar303: ahhh, forgot about that

kumar303: still, I don't think we'll be posting a JSON object as the request, will we?

andymckay: if you POST application/json, then yeah

kumar303:
I think you should document the ``__braintree__`` error responses with an example of what an error response will look like

PR #382, opened 2015-04-28
add in payments service key (fixes #372)

### LINKED ISSUE #372: Add in payments-service oauth keys
There are oauth keys for marketplace to do auth (using curling). We should add in keys for payments-service so it can do the same - but we distinguish the caller.

andymckay:


kumar303:
Sweet. Don't we have a way now to make it work out of the box in Docker though?

r+

PR #384, opened 2015-04-29
add in braintree buyer and fix up the mocks

andymckay:


lib/brains/tests/base.py
```diff
@@ -1,18 +1,83 @@
...
+    """
+    A standard APITest with the ability to provide Braintree mocks.
+
+    Add in dict of gateways to override onto the class, for example:
+
+            gateways = {'client': ClientTokenGateway}
+
+    Access the mock lke this:
+
+            self.mocks['client'].generate.return_value = 'a-sample-token'
```
kumar303: I know it's more letters to type but I think this would make more sense as:

    self.mocks['ClientTokenGateway'].generate.return_value

you wouldn't need a mapping like that either

andymckay: i'm trying to avoid always typing in gateway because it makes me sad

kumar303: oh, I do like how the mapping is defined at the top of each test

kumar303:
looks like maybe a legit test failure

lib/brains/models.py
```diff
@@ -0,0 +1,28 @@
...
+from lib.buyers.models import Buyer
+from solitude.base import Model
+
+
+class BraintreeBuyer(Model):
+
+    """A holder for any braintree specific stuff around the buyer."""
+    # Allow us to turn off braintree buyers if we need to.
+    active = models.BooleanField(default=True)
+    # The specific braintree-id confirming to braintree's requirements.
```
kumar303: typo: conforming

lib/brains/tests/base.py
```diff
@@ -1,18 +1,83 @@
...
-        if client.mocks:
-            raise MockError('Unused braintree client mock exists: {0}'
-                            .format(client.mocks))
-        self.set_mocks()
-        super(BraintreeTest, self).tearDown()
+        self.classes = {}
+        self.mocks = {}
+
+        for key, gateway in self.gateways.items():
+            name = gateway.__name__
```
kumar303: Does gateway have to be an object? I guess that forces each test to import it thus avoiding typos? Seems like ``{'client': 'ClientTokenGateway'}`` could be enough though.

lib/brains/tests/base.py
```diff
@@ -1,18 +1,83 @@
...
+
+            # Patch all the clasesses in the gateway.
```
kumar303: so many classes (typo)

lib/brains/tests/test_buyer.py
```diff
@@ -0,0 +1,47 @@
...
+from django.core.urlresolvers import reverse
+
+from nose.tools import eq_
+
+from lib.brains.models import BraintreeBuyer
+from lib.brains.tests.base import BraintreeTest
+from lib.buyers.models import Buyer
+
+
+class TestCustomer(BraintreeTest):
```
kumar303: does this need to inherit from ``BraintreeTest`` ? I guess it doesn't hurt.

lib/brains/tests/test_client.py
```diff
@@ -1,29 +1,12 @@
...
 
 class TestClient(BraintreeTest):
 
-    def test_mock_client(self):
-        assert isinstance(
-            get_client().Configuration.instantiate()._http_strategy,
-            HttpMock)
-
     def test_normal(self):
         with self.settings(BRAINTREE_MOCK=False,
```
kumar303: I guess you don't need ``BRAINTREE_MOCK``, right?

lib/brains/tests/test_customer.py
```diff
@@ -1,72 +1,88 @@
...
-        res = self.client.post(self.url, data={'uuid': 'minimal'})
+        self.mocks['customer'].find.return_value = customer()
+
+        res = self.client.post(self.url, data={'uuid': 'customer-id'})
         eq_(res.status_code, 201)
         buyer = Buyer.objects.get()
-        eq_(buyer.uuid, 'minimal')
+        eq_(buyer.uuid, 'customer-id')
+        braintree_buyer = BraintreeBuyer.objects.get()
+        eq_(braintree_buyer.braintree_id, 'test-{0}'.format(buyer.pk))
```
kumar303: The code under test needs to post to the braintree API, right? It seems risky to rely on the base test case teardown to verify that. It might be good to put an explicit assertion here that the braintree create method was called with the right parameters.

lib/brains/tests/test_token.py
```diff
@@ -1,13 +1,15 @@
...
 
 from lib.brains.tests.base import BraintreeTest
 
 
 class TestToken(BraintreeTest):
+    gateways = {'client': ClientTokenGateway}
 
     def test_token(self):
-        self.set_mocks(['POST', 'client_token', 200, 'token'])
+        self.mocks['client'].generate.return_value = 'a-sample-token'
```
kumar303: This might read better as ``mocks['token']``

lib/brains/views/customer.py
```diff
@@ -22,12 +23,27 @@ def create(request):
...
+    # Create the solitude objects.
+    buyer, buyer_created = Buyer.objects.get_or_create(uuid=uuid)
+    log.info('Buyer {0}: {1}'
+             .format('created' if buyer_created else 'exists', buyer.pk))
+
+    braintree, braintree_buyer_created = (
+        BraintreeBuyer.objects.get_or_create(buyer=buyer))
+    log.info('BraintreeBuyer {0}: {1}'
+             .format('created' if braintree_buyer_created else 'exists',
+                     braintree.pk))
```
kumar303: I don't like how the BraintreeBuyer object is saved to the database before one is created using the braintree API. This means that the API call fails, the database will be out of sync (I don't see any db rollback here). What benefit does this approach give us? Could you just wait until after the API call to save to the database? Otherwise, I think we'd need a new model field like ``exists_in_api=BooleanField()`` so that we can do lookups. That might make things more complicated though.

migrations/54-add-braintree-buyer.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE `buyer_braintree` (
+    `id` int(11) unsigned AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime NOT NULL,
+    `modified` datetime NOT NULL,
+    `counter` bigint,
+    `active` bool NOT NULL,
```
kumar303: does this need the default value? I forget how it works in django

kumar303:
I'm r- on saving to the DB before the customer exists in the braintree API but maybe I'm missing something

lib/brains/views/customer.py
andymckay: Its all in a transaction when it fails the whole transaction is rolled back because solitude has transactions from the middleware.

lib/brains/tests/test_customer.py
andymckay: Seems silly to write code to test it happens and then not trust it, I'm cool with checking a specific value, but don't want to have to write stuff that is already checked elsewhere.

solitude/exceptions.py
```diff
@@ -4,8 +4,17 @@
...
+from django.db.transaction import rollback
+
+from solitude.logger import getLogger
+
+log = getLogger('s')
+
 
 def custom_exception_handler(exc):
+    log.warning('Transaction rolled back due to error.')
+    rollback()
```
kumar303: I don't know if this will do what you want (because I'm not sure if transactions are enabled like this for mysql). Also, it doesn't seem to be testable probably because custom error handlers aren't triggered from the test client, I think. What if we turned on per request transactions? https://docs.djangoproject.com/en/1.8/topics/db/transactions/#tying-transactions-to-http-requests

lib/brains/views/customer.py
kumar303: OK, I see. I'm suspicious that this is not testable though (as you found in your SkipTest). Are you sure it's configured right?

kumar303:
looks like we can't test the rollback in the exception handler so that's a bummer. Oh well.

r+wc

PR #385, opened 2015-04-30
live tests with some oauth action

andymckay:
Also fixes 403 handling, which was previously throwing a 500 error, not a 403.

Secrets are not available on pull requests, so the live tests will fail until we merge this.

### DIFF
--- a/solitude/settings/base.py
+++ b/solitude/settings/base.py
@@ -355,8 +355,17 @@ NOSE_ARGS = [
     '--logging-filter=xsd',
     '--with-nicedots',
     '--with-blockage',
-    '--http-whitelist=""',
 ]
+
+if not os.environ.get('LIVE_TESTS'):
+    NOSE_ARGS += [
+        '-a !live',
+        '--http-whitelist=']
+else:
+    NOSE_ARGS += [
+        '-a live',
+        '--http-whitelist=localhost,api.sandbox.braintreegateway.com'
+    ]
 
 # Below is configuration of payment providers.
 
--- a/solitude/settings/test.py
+++ b/solitude/settings/test.py
@@ -62,6 +62,11 @@ CHECK_BANGO_TOKEN = True
 
 REQUIRE_OAUTH = False
 
+# Live server tests require this, otherwise they will fail. There is no static
+# content on our site, so meh.
+STATIC_URL = '/'
+STATIC_ROOT = '/'
+
 
 # Suds keeps a cache of the WSDL around, so after completing the test run,
 # lets remove that so it doesn't affect the next test run.
--- a/solitude/tests/live.py
+++ b/solitude/tests/live.py
@@ -1,0 +1,40 @@
+import os
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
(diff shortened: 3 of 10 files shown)


solitude/settings/base.py
```diff
@@ -355,9 +355,18 @@
     '--logging-filter=xsd',
     '--with-nicedots',
     '--with-blockage',
-    '--http-whitelist=""',
 ]
 
+if not os.environ.get('LIVE_TESTS'):
```
kumar303: I don't think you can simply bool the value. Consider this:
````
>>> bool("0")
True
````

Comment (kumar303) [inline solitude/settings/base.py]: but maybe that's ok. We wouldn't want the code to get too ugly.

File: solitude/tests/live.py [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+}
+for key in envs.keys():
+    value = ''
+    if local:
+        log.info('Found {0} from local.py.'.format(key))
+        value = getattr(local, key)
+
+    if not value:
+        log.info('Found {0} from environment.'.format(key))
+        value = os.environ.get(key)
```
kumar303: Why is this necessary? The [base settings](https://github.com/andymckay/solitude/blob/378/solitude/settings/base.py) file already gets values from the environment like ``BRAINTREE_MERCHANT_ID = os.getenv('BRAINTREE_MERCHANT_ID', '')``

solitude/settings/test.py
```diff
@@ -55,6 +55,10 @@
 
 BOKU_SECRET_KEY = 'for-testing-only'
 
+BRAINTREE_MERCHANT_ID = 'testing-merchant-id'
+BRAINTREE_PUBLIC_KEY = 'testing-public-key'
+BRAINTREE_PRIVATE_KEY = 'testing-private-key'
```
kumar303: Why do these need to be set?

solitude/tests/live.py
kumar303: In other words, if the value is not defined in a *local* settings file then you could just use the base value

kumar303: I suppose the ``settings_test.py`` values are the cause of why you have to look for environment variables again. What if ``settings_test.py`` didn't override anything? We have nose blockage.

solitude/settings/base.py
andymckay: We could test specific values, but testing for the presence of the env is enough and pretty common.

solitude/settings/test.py
andymckay: Can't spot a reason now, will remove.

solitude/tests/live.py
andymckay: I can't find a reason now either. Will remove.

solitude/tests/live.py
```diff
@@ -0,0 +1,40 @@
...
+
+from nose.plugins.attrib import attr
+
+from curling import lib
+
+from solitude.base import getLogger
+
+log = getLogger('s.tests')
+
+configs = {
```
kumar303: could call this ``overrides`` too

kumar303:
looks good, r+

PR #386, opened 2015-04-30
Link to braintree settings in docs

kumar303:
Also:
- install docs modules in docker image
- show the RTD theme when building docs locally


andymckay:
r+

PR #387, opened 2015-04-30
Preserve docker bash history

kumar303:


Dockerfile
```diff
@@ -5,6 +5,7 @@
 FROM mozillamarketplace/centos-mysql-mkt:0.2
 
 RUN yum install -y supervisor
+RUN yum install -y bash-completion
```
andymckay: As @muffinresearch once said, we should be pushing this up a bit into our general stack. I didn't do that for supervisor, but I thought I'd pass on the favour.

andymckay:
r+

Dockerfile
kumar303: ah, yeah, move to the base image? That would be cool

Issue #388, opened 2015-04-30
Upgrade transaction handling


andymckay-limited-access:
Upgrade to Django 1.8, remove the now deprecated TransactionMiddleware, start using the new Atomic code.

Then check that all the places that are aborting the transaction are.

Then (hopefully) write a unit test that allows us to actually test this decently.

PR #389, opened 2015-04-30
cope with being in an atomic block

andymckay:


andymckay:
Related to #388

andymckay:
This is wrong, the problem is (I think) that mocking out rollback barfs the transaction for the rest of the tests.

PR #392, opened 2015-05-01
create a payment method api and model

andymckay:
Fixes mozilla/payments#17

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -61,6 +61,8 @@ customer exists in Braintree already, it is not created.
 .. note:: If the creation fails at Braintree then the new buyer objects will
           not be created.
 
+.. _braintree-buyer-label:
+
 Braintree Buyer
 ---------------
 
@@ -98,4 +100,90 @@ object and Braintree customer.
 .. http:get:: /braintree/buyer/
 
     :query buyer: the primary key of the buyer.
+    :query active: the active status.
+
+Create a payment method
+-----------------------
+
+Creates a payment method in Braintree and the corresponding payment method in
+solitude.
+
+.. http:post:: /braintree/paymethod/
+
+    :<json string uuid: the uuid of the buyer in solitude.
+    :<json string nonce: the payment nonce returned by Braintree.
+
+    Returns :ref:`a payment method object <payment-methods>` with some extra
+    fields.
+
+    .. code-block:: json
+
+        {
+            "braintree": {
+                "created_at": "2015-05-05T14:22:26.650",
+                "token": "da-token",
+                "updated_at": "2015-05-05T14:22:26.650"
+            },
+            "mozilla": {
+                "active": true,
+                "braintree_buyer": "/braintree/mozilla/buyer/16/",
+                "counter": 0,
+                "created": "2015-05-05T14:22:26.656",
+                "id": 4,
+                "modified": "2015-05-05T14:22:26.656",
+                "provider_id": "da-token",
+                "resource_pk": 4,
+                "resource_uri": "/braintree/mozilla/paymethod/4/",
+                "truncated_id": "7890",
+                "type": 1,
+                "type_name": "visa"
+            }
+        }
+
+    :>json string braintree token: id of the payment method in braintree.
+    :>json string braintree created_at: created date and time.
+    :>json string braintree updated_at: updated date and time.
+
+    :status 201: payment method created.
+
+Data stored in solitude
+-----------------------
+
+Information about the payment method is stored in solitude.
+
+.. http:get:: /braintree/mozilla/paymethod/<method id>/
+
+    .. code-block:: json
+
+        {
+          "active": true,
+          "braintree_buyer": "/braintree/mozilla/buyer/2/",
+          "counter": 0,
+          "created": "2015-05-05T14:25:38",
+          "id": 1,
+          "modified": "2015-05-05T14:25:38",
+          "provider_id": "da-token",
+          "resource_pk": 1,
+          "resource_uri": "/braintree/mozilla/paymethod/1/",
+          "truncated_id": "some",
+          "type": 1,
+          "type_name": "visa"
+        }
+
+    :>json boolean active: active flag for the method.
+    :>json string braintree_buyer: URI to :ref:`a braintree buyer object <braintree-buyer-label>`.
+    :>json string provider_id: an id for the payment method on the provider, this field is read only.
+    :>json string truncated_id: a truncated id of the payment type, for example
+                                for a credit card, the last 4 digits, this field is read only.
+    :>json int type: `1` for credit card is currently the only one supported, this field is read only.
+    :>json string type_name: name of the type of purchase, this field is read only.
+
+.. http:patch:: /braintree/mozilla/<method id>/
+
+    :<json boolean active: if the buyer is currently active.
+
+.. http:get:: /braintree/mozilla/
+
+    :query braintree_buyer: the primary key of the braintree_buyer.
+    :query braintree_buyer__buyer__uuid: the uuid for the buyer.
     :query active: the active status.
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -1,5 +1,39 @@
 from django import forms
+from django.core.exceptions import ObjectDoesNotExist
+
+from lib.buyers.models import Buyer
 
 
 class BuyerForm(forms.Form):
     uuid = forms.CharField(max_length=255)
+
+
+class PaymentMethodForm(forms.Form):
+    buyer_uuid = forms.CharField(max_length=255)
+    nonce = forms.CharField(max_length=255)
+
+    def clean_buyer_uuid(self):
+        data = self.cleaned_data
+
+        try:
+            self.buyer = Buyer.objects.get(uuid=data['buyer_uuid'])
+        except ObjectDoesNotExist:
+            raise forms.ValidationError('Buyer does not exist.')
+
+        try:
+            self.braintree_buyer = self.buyer.braintreebuyer
+        except ObjectDoesNotExist:
+            raise forms.ValidationError('Braintree buyer does not exist.')
+
+        return data
+
+    @property
+    def braintree_data(self):
+        return {
+            'customer_id': str(self.braintree_buyer.braintree_id),
+            'payment_method_nonce': self.cleaned_data['nonce'],
+            # This will force the card to be verified upon creation.
+            'options': {
+                'verify_card': True
+            }
+        }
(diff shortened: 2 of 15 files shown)


docs/topics/braintree.rst
```diff
@@ -57,3 +57,46 @@ Braintree already, it is not created.
...
+Create a payment method
+-----------------------
+
+Creates a payment method in Braintree and the corresponding payment method in
+solitude.
+
+.. http:post:: /braintree/method/
+
+    :<json string uuid: the uuid of the buyer in solitude.
+    :<json string nonce: the payment nonce returned by Braintree.
```
kumar303: I was trying to find something in brainree's docs that we can link to but I can't actually find it now :(

docs/topics/braintree.rst
```diff
@@ -57,3 +57,46 @@ Braintree already, it is not created.
...
+            "provider": 4,
+            "provider_id": "da-token",
+            "resource_pk": 1,
+            "resource_uri": "/generic/method/1/",
+            "truncated_id": "7890",
+            "type": 1,
+            "type_name": "visa"
+        }
+
+   :>json string braintree token: id of the payment method in braintree.
```
kumar303: I've been listing nested fields like ``braintree.token`` since that would be valid JavaScript. I wish there was a better way to display nested fields though.

lib/buyers/urls.py
```diff
@@ -6,6 +6,7 @@
 
 router = DefaultRouter()
 router.register(r'buyer', views.BuyerViewSet)
+router.register(r'method', views.PaymentMethodViewSet, base_name='method')
```
kumar303: I think ``paymethod`` would make this more clear. Since it's attached to the generic API, payments are not necessarily implied.

lib/brains/urls.py
```diff
@@ -12,5 +12,6 @@
     url(r'', include(router.urls)),
     url(r'^token/generate/$', 'token.generate', name='token.generate'),
     url(r'^customer/$', 'customer.create', name='customer'),
+    url(r'^method/$', 'method.create', name='method'),
```
kumar303: here also I think ``paymethod`` would be less ambiguous

lib/brains/views/method.py
```diff
@@ -0,0 +1,51 @@
...
+    PaymentMethodSerializer as BraintreeSerializer)
+from lib.buyers.models import PaymentMethod
+from lib.buyers.serializers import PaymentMethodSerializer
+from lib.transactions.constants import PROVIDER_BRAINTREE
+from solitude.constants import PAYMENT_METHOD_CARD
+from solitude.logger import getLogger
+
+
+log = getLogger('s.brains')
+
```
kumar303: Why can't you just define a GET endpoint here that queries ``PaymentMethod``? I think it would avoid confusion since it's weird that you can't post to the generic PaymentMethod endpoint. I also think it's perhaps overly optimistic that we will use the concept of payment methods for other providers.

lib/brains/forms.py
```diff
@@ -1,5 +1,40 @@
...
+    buyer_uuid = forms.CharField(max_length=255)
+    nonce = forms.CharField(max_length=255)
+
+    def clean_buyer_uuid(self):
+        data = self.cleaned_data
+
+        try:
+            self.buyer = Buyer.objects.get(uuid=data['buyer_uuid'])
+        except ObjectDoesNotExist:
+            raise forms.ValidationError('Buyer does not exist.')
```
kumar303: mmm, we need to figure out localization soon. We started doing it by raising mappable constant strings like https://github.com/mozilla/webpay/blob/master/webpay/pay/forms.py#L57 . I think that's the right way to approach it but we could punt this until later I suppose.

lib/brains/forms.py
```diff
@@ -1,5 +1,40 @@
...
+            raise forms.ValidationError('Braintree buyer does not exist.')
+
+        return data
+
+    @property
+    def braintree_data(self):
+        return {
+            'customer_id': str(self.braintreebuyer.braintree_id),
+            'payment_method_nonce': self.cleaned_data['nonce'],
+            'token': str(self.buyer.pk),
```
kumar303: this doesn't look right. I think it should be the primary key of the payment method object saved in our database, not the pk of the buyer.

lib/brains/views/method.py
andymckay: Because this way lies madness and some of the craziness that occurs in Bango and Boku where endpoints become the combination of two competing needs and functionalities. For example a serializer in DRF is used to read the data and output the data, its also pretty closely tied to a model and a serializer. If you get an end point that does multiple things you get some of the methods like Bango which currently have a form to process Solitude data and then a serializer to get Bango data and it gets confusing as hell.

After learning from that this feels really clean to me. You've got a way of querying solitude easily. But you can't change much from that, because really its controlled by the provider.

If we did get payment methods for Bango, we'd just add in a similar method for Bango and we'd be done. This approach lets us write a really tailored view for what Braintree needs and make it nice and clean for Payments for FxA. Something that you can see we suffered from in webpay.

Ok long answer, sorry.

lib/brains/forms.py
kumar303: or am I reading this wrong? https://developers.braintreepayments.com/javascript+python/reference/request/payment-method/create#token

lib/brains/tests/test_method.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def method(**kw):
+    method = {
+        'token': 'customer-id',
+        'created_at': datetime.now(),
+        'updated_at': datetime.now(),
+        'last_4': '7890',
+        'card_type': 'visa',
+        'token': 'da-token',
```
kumar303: token is defined multiple times

lib/brains/views/method.py
kumar303: ok, I understand it better. I guess to keep them fully separate we'd want to have the client post first to ``braintree/paymethod`` then secondly to ``generic/paymethod`` but that would be an extra post and extra failure management for the client. In other words, now it seems awkward to me that the braintree endpoint creates a generic payment method behind the scenes ;) but I guess it's good encapsulation.

kumar303:
I think the token parameter to braintree's API needs fixing, other than that I'm r+wc

lib/brains/views/method.py
andymckay: Now you've got me confused :) You've got one endpoint, which is a specialized endpoint for your needs that does all the work you need.

andymckay: Let's chat Monday. Its Friday, my brain is tired.

docs/topics/braintree.rst
andymckay: Works for me.

lib/brains/forms.py
andymckay: It shouldn't be the buyer.pk, because it won't be unique, so you are right there. I was originally thinking that if i make all the token/provider_id thing match the pk of an object it would make life that bit easier and maybe there's some crazy disaster recovery plan we could do where we could match things up.

But thinking about it, it really doesn't make much sense. I'm going to change these apis to not supply a token, let braintree provide one and store that - less code on our end, less worrying about one which pk it is.

lib/brains/tests/test_method.py
andymckay: thanks

lib/brains/forms.py
andymckay: It really depends where those values are going to be shown. I've looked at the ones in solitude and I don't think a single one of those will ever go up to an end user. I think they occurred during development and things were confusing and it became it stop gap.

The problem is I think that approach has to be applied to everything or nothing.

Perhaps there's other ways of doing this and it really depends where the logic is going to occur. It feels like its going to be in payments-service again and what happens there will be interesting.

andymckay: One example. No matching string for SELLER_UUID not being unique or any other errors on that form. 

https://github.com/mozilla/solitude/search?utf8=%E2%9C%93&q=BUYER_UUID_ALREADY_EXISTS&type=Code

docs/topics/braintree.rst
```diff
@@ -99,3 +101,89 @@ object and Braintree customer.
...
+          "resource_pk": 1,
+          "resource_uri": "/braintree/mozilla/paymethod/1/",
+          "truncated_id": "some",
+          "type": 1,
+          "type_name": "visa"
+        }
+
+    :>json boolean active: active flag for the method.
+    :>json string braintree_buyer: URI to :ref:`a braintree buyer object <braintree-buyer-label>`.
+    :>json int provider_id: an id for the payment method on the provider, this field is read only.
```
kumar303: this looks like a string (not an int) in the example

andymckay: good catch, thanks

lib/brains/urls.py
```diff
@@ -2,15 +2,17 @@
 
 from rest_framework.routers import DefaultRouter
 
-from lib.brains.views import buyer
+from lib.brains.views import buyer, paymethod
 
 router = DefaultRouter()
 router.register(r'buyer', buyer.BraintreeBuyerViewSet, base_name='buyer')
```
kumar303: I really liked how you split ``brains.views.*`` into specific modules that matched the URLs closely. Could things like ``mozilla/buyer`` get imported from ``lib.brains.views.mozilla.buyer`` ?

lib/brains/models.py
```diff
@@ -20,9 +21,35 @@ class Meta(Model.Meta):
...
+    provider_id = models.CharField(max_length=255)
+    # The type of payment method eg: card, paypal or bitcon
+    type = models.PositiveIntegerField(choices=(
+        (PAYMENT_METHOD_CARD, PAYMENT_METHOD_CARD),
+    ))
+    # Details about the type, eg Amex, Orange.
+    type_name = models.CharField(max_length=255)
+    # For credit cards, this is the last 4 numbers, could be a truncated
+    # phone number or paypal account for example.
+    truncated_id = models.CharField(max_length=255)
```
kumar303: this doesn't define the ``counter`` column. Do we need that column? If so, could you define it and add a comment about it what it means?

kumar303:
looks really good!

r+wc

lib/brains/models.py
andymckay: Counter is on the base model, along with created, updated and the like: https://github.com/mozilla/solitude/blob/master/solitude/base.py#L279. I haven't figured out why we need it yet. I think its something to do with etags.

lib/brains/urls.py
andymckay: Yeah, then I'd feel like I had to move the tests and serializers around as well and someone changes the URL structure and all the folders are wrong and... sound ok, but feels a bit fiddy.

PR #396, opened 2015-05-04
remove memcache fixes #395

### LINKED ISSUE #395: Remove memcache from solitude
The only place its used is in services, I think we can just remove it.

andymckay:


andymckay:
This isn't active on dev or stage, so not sure why it was turned on for docker. Rather than adding in another docker container, let's use same as dev and stage. Stage:

```
[amckay@web14.stage.addons.phx1 solitude]$ ../venv/bin/python manage.py shell
Python 2.7.5 (default, Sep 12 2013, 19:47:54) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-3)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from django.conf import settings
>>> settings.CACHES
{'default': {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'}}
```

solitude/settings/base.py
```diff
@@ -19,14 +19,6 @@
...
 ALLOWED_HOSTS = []
 
-if os.environ.get('MEMCACHE_URL'):
-    CACHES = {
-        'default': {
-            'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-            'LOCATION': os.getenv('MEMCACHE_URL'),
-        }
-    }
-
```
kumar303: I think we should leave this in just in case someone needs to reproduce a memcache related issue locally. Also, a quick look at prod settings tells me that it would rely on this env detection if it were to run with memcache.

kumar303:
Good idea. I noticed this in the status report too.

r+wc

solitude/settings/base.py
andymckay: Can you point where you looked please? I couldn't find anything else referring to memcache for solitude. I found that stage does not use it.

kumar303: Oh, we don't use memcache in prod at all? If that's the case then disregard my comment. What I meant was that if prod is using memcache then it must be doing so via the environment variable since it's not explicitly doing so in the settings.

andymckay: Yeah, and its confirmed with ops that we don't use memcache their either.

PR #401, opened 2015-05-06
start of clean up error handling

andymckay:


andymckay:
fixes #349

docs/topics/setup.rst
```diff
@@ -61,9 +61,13 @@ To run unit tests::
...
 
-Live server and Braintree integration tests can be run by::
+All live server and Braintree integration tests are not run by default. To run
+live server and Braintree integration tests::
 
-    LIVE_TESTS=yes python manage.py test
+    LIVE_TESTS=live,braintree python manage.py test
+
+The value of LIVE_TESTS is passed to the `nose args command
+<http://nose.readthedocs.org/en/latest/plugins/attrib.html#simple-syntax>`_.
```
kumar303: can links be broken across new lines like this?

lib/brains/tests/base.py
```diff
@@ -91,3 +96,16 @@ def cleanUpBrains(self):
                 if not getattr(v, call).called:
                     raise MockError('{0}.{1} registered but not called.'
                                     .format(self.gateways[k].__name__, call))
+
+
+@attr('braintree')
```
kumar303: This would also work if you need it: ``@attr('live', 'braintree')``

solitude/settings/base.py
```diff
@@ -349,15 +349,22 @@
...
-    NOSE_ARGS += [
-        '-a live',
-        '--http-whitelist=localhost,api.sandbox.braintreegateway.com'
-    ]
+    # If you specified something, just pass it through.
+    NOSE_ARGS.append('-a ' + tests)
+
+# Figure out the whitelist.
+http = 'localhost' if 'live' in tests else ''
+http += ',api.sandbox.braintreegateway.com' if 'braintree' in tests else ''
```
kumar303: this looks like you could end up with ``--http-whitelist=,api.sandbox.braintreegateway.com`` (a preceding comma). That seems like it could break something. You could work around it by making ``http`` a list and doing ``",".join(http)``

kumar303: Actually, yeah, a list would be cleaner:
````
http = []
if 'live' in tests:
    http.append('localhost')
...
````

File: solitude/tests/test_error.py [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+    def test_403_html(self):
+        with self.assertRaises(HttpClientError) as error:
+            lib.API(self.live_server_url).by_url('/this/does-not-exist/').get()
+        eq_(error.exception.response.json, {})
+
+    def test_drf_403_html(self):
+        with self.assertRaises(HttpClientError) as error:
+            lib.API(self.live_server_url).by_url('/generic/transaction/').get()
+        eq_(error.exception.response.json,
+            {'detail': 'Incorrect authentication credentials.'})
```
kumar303: ooh, neat.

solitude/views.py
```diff
@@ -1,28 +1,38 @@
...
+    # If returning JSON, then we can't send back an empty body, unless we
+    # return a 204 - the client should be able to parse it.
+    message = data if data else {}
+    if 'application/json' in request.META.get('HTTP_ACCEPT'):
+        return JsonResponse(message, status=status)
 
-    response = render_to_response('error.html')
-    response.status_code = status
-    return response
+    # If you send back something flasy dictionary with HTTPResponse, it just
```
kumar303: I'm not sure what this sentence means, can you re-word it?

solitude/views.py
```diff
@@ -1,28 +1,38 @@
...
+    # returns an empty body.
```
kumar303: ``message.values()`` does not look like an empty response. Is it?

kumar303:
The [Errors](https://raw.githubusercontent.com/andymckay/solitude/consistent-errors/docs/topics/overall.rst) section still looks like a placeholder. Can you add some examples to it?

solitude/views.py
```diff
@@ -1,28 +1,38 @@
...
-    # If they posted JSON, return the error as JSON, assuming that accept
-    # is set correctly.
-    if request.META['CONTENT_TYPE'] == 'application/json':
-        return handle_500(request, sys.exc_info()[1])
+def response(request, status=500, data=None):
+    # If returning JSON, then we can't send back an empty body, unless we
+    # return a 204 - the client should be able to parse it.
+    message = data if data else {}
+    if 'application/json' in request.META.get('HTTP_ACCEPT'):
+        return JsonResponse(message, status=status)
```
kumar303: I know it's not very HTTPlike but I feel like we should always return a JSON response since we don't really support anything else.

solitude/views.py
kumar303: what I mean is that ``messages.values()`` might not be falsey if data was specified

andymckay:
Doesn't seem worth expanding docs until we've decided what the errors are. I've documented quite a few of them in the mailing list thread.

docs/topics/setup.rst
andymckay: yup, the part in the <..> is wrapped by the UI, but seems to work just fine

solitude/settings/base.py
andymckay: It could, will change.

andymckay: Also, this would not distinguish between `test` and `!test`, which is now fixed.

solitude/views.py
andymckay: Just send back JSON and if they can't accept it, :fried_shrimp:

kumar303:
ok, adding docs later is probably fine. I was just thinking something basic though, like ``{"detail": "this is the detail of the error"}``

r+wc

PR #402, opened 2015-05-06
update customer to the newer style and assume a lookup is going to oc…

### LINKED ISSUE #398: Update customer
Update the braintree customer API to not point to the generic API, but instead point to the braintree specific buyer.

Also while there, maybe stop creating a customer id and let braintree assign one?

andymckay:
…cur before calling fixes #398

### DIFF
--- a/docs/topics/overall.rst
+++ b/docs/topics/overall.rst
@@ -37,6 +37,15 @@ Common elements in some responses:
   return a value of `no_uri`, or similar, that's
   `a bug <https://github.com/mozilla/solitude/issues/380>`_.
 
+* `created` (datetime): when the object is created. Using the Django Rest
+  Framework format, `ECMA 262 <http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.15>`_.
+
+* `counter` (int): a counter that increments on each save of an object. This is
+  used for etag generation.
+
+* `modified` (datetime): when the object was last modified. Using the Django Rest
+  Framework format, `ECMA 262 <http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.15>`_.
+
 
 Errors
 ~~~~~~
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -1,11 +1,25 @@
 from django import forms
 from django.core.exceptions import ObjectDoesNotExist
 
+from lib.brains.models import BraintreeBuyer
 from lib.buyers.models import Buyer
 
 
 class BuyerForm(forms.Form):
     uuid = forms.CharField(max_length=255)
+
+    def clean_uuid(self):
+        data = self.cleaned_data
+
+        try:
+            self.buyer = Buyer.objects.get(uuid=data['uuid'])
+        except ObjectDoesNotExist:
+            raise forms.ValidationError('Buyer does not exist.')
+
+        if BraintreeBuyer.objects.filter(buyer=self.buyer).exists():
+            raise forms.ValidationError('Braintree buyer already exists.')
+
+        return data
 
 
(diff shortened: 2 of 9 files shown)


kumar303:
I think sticking to [50 char first line commit messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) is still a good idea. It seems that all of the tooling (github, ``git log``, etc) is optimized for it.

docs/topics/overall.rst
```diff
@@ -37,6 +37,15 @@ Common elements in some responses:
...
   `a bug <https://github.com/mozilla/solitude/issues/380>`_.
 
+* `created` (datetime): when the object is created. Using the Django Rest
+  Framework format, `ECMA 262 <http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.15>`_.
+
+* `counter` (int): a counter that increments on each save of an object. This is
+  used for etag generation.
+
+* `modified` (datetime): when the object was last modified. Using the Django Rest
+  Framework format, `ECMA 262 <http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.15>`_.
```
kumar303: You could make ```ECMA 262`_`` into a shared footnote link

lib/brains/errors.py
```diff
@@ -6,6 +6,13 @@ class MockError(Exception):
     """
 
 
+class BraintreeError(Exception):
```
kumar303: I don't see this used anywhere

lib/brains/errors.py
```diff
@@ -6,6 +6,13 @@ class MockError(Exception):
...
+
+    """
+    A standard Braintree error.
+    """
+
+
 class BraintreeResultError(Exception):
```
kumar303: Should this inherit from ``BraintreeError`` ?

lib/brains/forms.py
```diff
@@ -1,12 +1,26 @@
...
 class BuyerForm(forms.Form):
     uuid = forms.CharField(max_length=255)
 
+    def clean_uuid(self):
+        data = self.cleaned_data
+
+        try:
+            self.buyer = Buyer.objects.get(uuid=data['uuid'])
+        except ObjectDoesNotExist:
+            raise forms.ValidationError('Buyer does not exist.')
```
kumar303: I don't think this is a good way to return validation errors. I filed https://github.com/mozilla/solitude/issues/405

lib/brains/views/customer.py
```diff
@@ -22,38 +19,21 @@ def create(request):
...
+
+    log.info('Braintree customer created: {0}'.format(result.customer.id))
+
+    braintree_buyer = BraintreeBuyer.objects.create(
+        buyer=form.buyer, braintree_id=result.customer.id)
+
+    log.info('Braintree buyer created: {0}'.format(braintree_buyer.pk))
+
+    res = serializers.Namespaced(
+        serializers.LocalBuyer(instance=braintree_buyer),
```
kumar303: I think it would be nicer to read this code as:
````
res = serializers.Namespaced(
    braintree=...,
    mozilla=...,
)
````
You can pass keyword args for positional args so it will work.

kumar303:
I don't know what's going on with ``BraintreeError`` but I'm r+wc otherwise

lib/brains/errors.py
andymckay: BraintreeError should be removed, it was something I was playing with before I simplified.

lib/brains/forms.py
andymckay: See email to dev-payments about this. Or I can comment on the issue.

andymckay:
getting excited on the commit messsage!

PR #403, opened 2015-05-06
a meaner django filter

### LINKED ISSUE #364: Make django-filter meaner
In the bad old tasty pie tradition, if you get filters wrong in listing urls, DRF ignores them. Meaning that you can easily return all objects.

I had a patch to stop this in tasty pie, need to repeat for DRF.

andymckay:
fixes #364

### DIFF
--- a/lib/sellers/views.py
+++ b/lib/sellers/views.py
@@ -13,5 +13,6 @@ class SellerProductViewSet(NonDeleteModelViewSet):
     queryset = SellerProduct.objects.all()
     serializer_class = SellerProductSerializer
     filter_fields = (
-        'external_id', 'public_id', 'seller__uuid', 'seller__active'
+        'external_id', 'public_id', 'seller__uuid', 'seller__active',
+        'seller'
     )
--- a/solitude/exceptions.py
+++ b/solitude/exceptions.py
@@ -1,5 +1,7 @@
+from django.conf import settings
 from django.db.transaction import rollback
 
+from rest_framework.exceptions import ParseError
 from rest_framework.response import Response
 from rest_framework.views import exception_handler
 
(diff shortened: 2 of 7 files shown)


lib/sellers/views.py
```diff
@@ -13,5 +13,6 @@ class SellerProductViewSet(NonDeleteModelViewSet):
     queryset = SellerProduct.objects.all()
     serializer_class = SellerProductSerializer
     filter_fields = (
-        'external_id', 'public_id', 'seller__uuid', 'seller__active'
+        'external_id', 'public_id', 'seller__uuid', 'seller__active',
+        'seller'
```
andymckay: Kind of satisfying that writing this code found this one missing, but we'll need to check flows on -dev to see if there's anything webpay or zamboni hits that we haven't tested for.

solitude/exceptions.py
```diff
@@ -10,9 +12,17 @@
...
+
+
 def custom_exception_handler(exc):
-    log.warning('Transaction rolled back due to error.')
-    rollback()
+    # We shouldn't be rolling back if we are inside a transaction test case
+    # until we figure out mozilla/solitude#388
+    if not settings.IN_TEST_SUITE:
+        log.warning('Transaction rolled back due to error.')
+        rollback()
```
kumar303: maybe an ``else`` block here with a log warning that transaction rollback was skipped?

solitude/filter.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework.filters import DjangoFilterBackend
+
+from solitude.exceptions import InvalidQueryParams
+
+
+class MeanFilter(DjangoFilterBackend):
```
kumar303: I support cruelty when it comes to preventing programmer mistakes but maybe a better name would be ``StrictInputFilter`` or ``StrictQueryFilter``?

kumar303:
most excellent! I've been wanting this for a while. Thanks.

r+wc

kumar303:
this might be a good place to document some cons to a strict filter approach:

* the client can no longer do ``GET /thing/?_cacheBuster=1234``
* deprecating (yet supporting) old requests may mean maintaining older, no longer used query parameters in the whitelist when filters change

I don't think either of these are deal breakers. The cache thing doesn't really apply to Solitude anyway.

solitude/exceptions.py
andymckay: Which will only happen in tests which people won't see, i'll worry about when I get to #388

solitude/filter.py
andymckay: Sure :)

PR #407, opened 2015-05-08
upgrade django

andymckay:
* upgrade django and django-nose
* remove IN_TEST_SUITE
* move to ATOMIC_REQUESTS
* add in rollback that doesn't barf in tests
* add tests to braintree
* ensure that one wierd test case in Bango status still works


andymckay:
fixes #388

lib/bango/errors.py
```diff
@@ -53,3 +53,13 @@ class ProcessError(Exception):
 
     def __init__(self, response):
         self.response = response
+
+
+class StatusError(Exception):
```
kumar303: I don't see this used anywhere. Did I miss it?

solitude/exceptions.py
```diff
@@ -18,11 +17,8 @@ class InvalidQueryParams(ParseError):
 
 
 def custom_exception_handler(exc):
-    # We shouldn't be rolling back if we are inside a transaction test case
-    # until we figure out mozilla/solitude#388
-    if not settings.IN_TEST_SUITE:
-        log.warning('Transaction rolled back due to error.')
-        rollback()
+    if not getattr(exc, '_do_not_rollback', False):
```
kumar303: a log message would be nice here since it could happen in prod

kumar303: at least a log message on the else branch which is when exceptions won't cause rollbacks

lib/bango/errors.py
```diff
@@ -53,3 +56,13 @@ class ProcessError(Exception):
...
+
+
+class StatusError(ParseError):
+
+    """
+    This is an error that occurs in the Stats checking code, which we do want
+    to commit so we've got a history of the transaction failure.
+    """
+
+    _do_not_rollback = True
```
kumar303: I don't think this is the right way to solve the problem.

Here's the problem:
* you want to raise an exception that indicates a failure
* everything written to the database is now questionable and should be rolled back because there was an exception (I know this doesn't apply to stats but I think in other cases it would)
* you want to commit some changes to the database *about the error*

I think this would be a better approach:

````
class StatusError:

    def report_error(self):
        self.obj.status = STATUS_BAD
        self.obj.errors = json.dumps({'form.errors': self.form_errors})
        self.obj.save()
````

The error raising code:
````
if not form.is_valid():
    raise StatusError(obj, form.errors)
````

The final error catching code:
````
    if hasattr(exc, 'report_errors'):
        transaction.rollback()
        exc.report_errors()
        set_rollback(False)
    else:
        set_rollback(True)  # let Django roll back
````
Without something like this, you'd never be able to roll back database changes **and** report errors.

**Comment by kumar303** [comment]
r- with proposal for another error reporting approach but I'm open to ideas

File: lib/bango/errors.py (hunk 3 above)
Comment (andymckay) [inline lib/bango/errors.py]: That's a pretty neat idea. I don't think it applies to this view, but its probably better to set it up correctly the first time. I'll noodle on this for a little bit.

File: solitude/tests/test_error.py [hunk 4]
```diff
@@ -22,3 +26,18 @@ def test_drf_403_html(self):
             lib.API(self.live_server_url).by_url('/generic/transaction/').get()
         eq_(error.exception.response.json,
             {'detail': 'Incorrect authentication credentials.'})
+
+
+class Rollback(Exception):
+    _do_not_rollback = True
```
kumar303: it looks like you don't need any of this anymore, right?

kumar303:
r+wc

solitude/tests/test_error.py
andymckay: Good catch, thanks.

PR #410, opened 2015-05-11
logging fixes #409

### LINKED ISSUE #409: Fix logging
I added in a new row to the logs: status-reason, but didn't add it to the headers.

Let's fix that up and hopefully change it so it will be harder to do in the future.

And I'll need to regenerate a crap ton of solitude logs to fix that up.

andymckay:


lib/transactions/models.py
```diff
@@ -92,21 +94,23 @@ def is_refunded(self):
...
+            ('status', self.status),
+            ('reason', self.status_reason),
+            ('buyer', self.buyer.uuid if self.buyer else None),
+            ('seller', (self.seller_product.seller.uuid
+                        if self.seller_product else None)),
+            ('source', self.source),
+            ('carrier', self.carrier),
+            ('region', self.region),
+            ('provider', self.provider)
+        ))
```
robhudson: The `reason` is new here and may break stats. Is there a bug for that? Ideally we'd only append new columns to avoid breaking other things that rely on the order.  See https://github.com/mozilla/monolith-aggregator/blob/master/monolith/aggregator/plugins/solitude.py#L87

andymckay: Bad news, reason was added last month:

https://github.com/mozilla/solitude/commit/705ac69dd03638e081ee87b8b1ccc63cf8f9f0aa#diff-3d2b1aa083001b6a0456c9031c1ee15cR96

And I bumped the version then.

robhudson:
r+

Issue #359, opened 2015-04-21
Remove Boku


andymckay-limited-access:
Since Boku isn't happening, remove it from solitude.

PR #414, opened 2015-05-15
make braintree use solitude-auth

andymckay:
fixes mozilla/payments#19

### DIFF
--- a/bin/docker/supervisor.conf
+++ b/bin/docker/supervisor.conf
@@ -11,9 +11,14 @@ stdout_logfile=logs/docker.log
 stdout_logfile_maxbytes=1MB
 stopsignal=KILL
 environment=
+    SOLITUDE_BRAINTREE_PROXY="http://solitude-auth:2603/v1/braintree/",
+    # Note: that this proxy setting will be replaced when solitude-auth
+    # is completed for all providers.
     SOLITUDE_ZIPPY_PROXY="http://solitude:2603/proxy/provider",
     SOLITUDE_URL="http://solitude:2602"
 
+# Note: that this will be removed when solitude-auth is completed for all
+# providers.
 [program:proxy]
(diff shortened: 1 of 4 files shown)


andymckay:
fixes mozilla/solitude#359

bin/docker/supervisor.conf
```diff
@@ -11,9 +11,14 @@ stdout_logfile=logs/docker.log
 stdout_logfile_maxbytes=1MB
 stopsignal=KILL
 environment=
+    SOLITUDE_BRAINTREE_PROXY="http://solitude-auth:2603/v1/braintree/",
```
kumar303: I think it would be just as fine to call this ``SOLITUDE_BRAINTREE_AUTH`` - might as well fully commit to "auth" 100% ;)

kumar303: oh I see that this is a temporary variable maybe left over from the previous proxy approach. Okay.

lib/brains/client.py
```diff
@@ -8,11 +9,31 @@
...
 class Http(braintree.util.http.Http):
 
-    def http_do(self, *args, **kw):
+    def http_do(self, verb, path, headers, body):
+        # Tell solitude-auth where we really want this request to go to.
+        headers['x-solitude-url'] = self.environment._real.base_url + path
+        # Set the URL of the request to point to the auth server.
+        path = self.env
+
         with statsd.timer('solitude.braintree.api'):
```
kumar303: maybe this statsd tag should be changed to something with ``auth`` in it?

kumar303:
cool!

r+wc

lib/brains/client.py
andymckay: Yeah good idea. I think everything gets prefixed so its in a different namespace, but will help confusion later on.

Issue #417, opened 2015-05-15
Reformat Bango errors


andymckay-limited-access:
Make Bango errors use the new format and return 422 errors.

Issue #416, opened 2015-05-15
Upgrade DRF


andymckay-limited-access:
To version 3.x for a bunch of improvements, but mostly for the change in mixins which will make them raise serialization errors, instead of returning a 400 response.

This prevents all errors from being formatted consistently, which is mozilla/solitude#349

andymckay:
Duped to #348

PR #418, opened 2015-05-15
start of formatting errors more consistently

andymckay:
helps #349, but #416 and #417 are both needed to complete this

### DIFF
--- a/docs/topics/overall.rst
+++ b/docs/topics/overall.rst
@@ -46,5 +46,72 @@ Common elements in some responses:
 Errors
 ~~~~~~
 
-Errors. Consistent interface to be determined
+Errors. Consistent interface in progress:
 `by this issue <https://github.com/mozilla/solitude/issues/349>`_.
+
+To seperate the old and new style, two different kinds of errors will be returned
+a status of `400 <http://httpstatus.es/400>`_ for old format errors and
+`422 <http://httpstatus.es/422>`_ for new format errors.
+
+400
+---
+
+Responses are currently inconsistent and pending upon
+fixes to Bango and `upgrading to Django Rest Framework 3.x <https://github.com/mozilla/solitude/issues/416>`_.
+
+422
+---
+
+Errors will be raised with the namespace of the error, currently one of `mozilla`,
+`braintree` or `bango` to represent the part of the system that caused the error.
+
+Mozilla
+~~~~~~~
+Form errors in Solitude are given the namespace `mozilla`.
+
+An error contains the field the error occurred on and the message and code. It is
+possible for more than one error on a field. For a consistent interface use the code
+attribute.
+
+Example failure in form processing::
+
+    .. code:json::
+
+        {
+            "mozilla": {
+                "name": [
+                    {"message": "First error", 'code': "first"},
+                    {"message": "Second error", 'code': "second"}
+                ],
+                "__all__": [
+                    {"message': "Non field error", "code: "non-field"}
+                ]
+            }
+        }
+
+In this example `name` is a field passed in the request. The `__all__` refers
+to an error that did not exist on a field.
+
+Braintree
+~~~~~~~~~
+Data errors in Braintree are given the namespace `braintree`.
+
+An error contains the field the error occurred on and the message and code. It is
+possible for more than one error on a field. For a consistent interface use the code
+attribute, the code attribute is referenced in the
+`Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/general/validation-errors/all>`_
+
+Errors occur on Braintree fields, not fields passed in the request, so the the error
+keys do not match request fields.
+
+Example failure from Braintree::
+
+    .. code:json::
+
+        {
+            "braintree": {
+                "payment_method_token": [
+                    {'message': 'Payment method token is invalid.', 'code': '91903'},
+                ]
+            }
+        }
--- a/docs/topics/overall.rst
+++ b/docs/topics/overall.rst
@@ -46,7 +46,7 @@ Common elements in some responses:
 Errors
 ~~~~~~
 
-Errors. Consistent interface in progress:
+Errors. Consistent interface in progress and tracked 
 `by this issue <https://github.com/mozilla/solitude/issues/349>`_.
 
 To seperate the old and new style, two different kinds of errors will be returned
@@ -70,8 +70,8 @@ Mozilla
 Form errors in Solitude are given the namespace `mozilla`.
 
 An error contains the field the error occurred on and the message and code. It is
-possible for more than one error on a field. For a consistent interface use the code
-attribute.
+possible for more than one error to exist on a field. For a consistent interface
+use the `code` attribute.
 
 Example failure in form processing::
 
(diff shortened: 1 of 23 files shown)


docs/topics/overall.rst
```diff
@@ -46,5 +46,72 @@ Common elements in some responses:
 Errors
 ~~~~~~
 
-Errors. Consistent interface to be determined
+Errors. Consistent interface in progress:
 `by this issue <https://github.com/mozilla/solitude/issues/349>`_.
```
kumar303: 'by this issue' no longer makes sense in this sentence

docs/topics/overall.rst
```diff
@@ -46,5 +46,72 @@ Common elements in some responses:
...
+
+Errors will be raised with the namespace of the error, currently one of `mozilla`,
+`braintree` or `bango` to represent the part of the system that caused the error.
+
+Mozilla
+~~~~~~~
+Form errors in Solitude are given the namespace `mozilla`.
+
+An error contains the field the error occurred on and the message and code. It is
+possible for more than one error on a field. For a consistent interface use the code
```
kumar303: typo: ...possible for more than one error [to exist] on a field

docs/topics/overall.rst
```diff
@@ -46,5 +46,72 @@ Common elements in some responses:
...
+
+In this example `name` is a field passed in the request. The `__all__` refers
+to an error that did not exist on a field.
+
+Braintree
+~~~~~~~~~
+Data errors in Braintree are given the namespace `braintree`.
+
+An error contains the field the error occurred on and the message and code. It is
+possible for more than one error on a field. For a consistent interface use the code
```
kumar303: typo again: 'for more than one error...'

lib/brains/errors.py
```diff
@@ -6,12 +11,27 @@ class MockError(Exception):
...
+                'message': error.message
+            })
+
+        return {'braintree': dict(errors)}
+
+
 class BraintreeResultError(Exception):
 
     """
     When an error occurs in the result that is not a standard error.
```
kumar303: should be a comma at the end of this sentence

lib/brains/forms.py
```diff
@@ -16,10 +16,12 @@ def clean_uuid(self):
         try:
             self.buyer = Buyer.objects.get(uuid=data['uuid'])
         except ObjectDoesNotExist:
-            raise forms.ValidationError('Buyer does not exist.')
+            raise forms.ValidationError('Buyer does not exist.',
+                                        code='does_not_exist')
```
kumar303: I think ``buyer_does_not_exist`` would be less ambiguous because there are other entities at play such as sellers and products

lib/brains/forms.py
```diff
@@ -16,10 +16,12 @@ def clean_uuid(self):
...
 
         if BraintreeBuyer.objects.filter(buyer=self.buyer).exists():
-            raise forms.ValidationError('Braintree buyer already exists.')
+            raise forms.ValidationError('Braintree buyer already exists.',
+                                        code='already_exists')
```
kumar303: same here: ``buyer_already_exists``

lib/brains/forms.py
```diff
@@ -61,13 +65,14 @@ class SubscriptionForm(forms.Form):
...
 
         try:
             obj = SellerProduct.objects.get(external_id=data['plan'])
         except ObjectDoesNotExist:
-            raise forms.ValidationError('Seller product does not exist.')
+            raise forms.ValidationError(
+                'Seller product does not exist.', code='does_not_exist')
 
         self.seller_product = obj
         return data
```
kumar303: since you switched to ``clean_<field>`` you need to return the plan value, no? From the [docs](https://docs.djangoproject.com/en/1.7/ref/forms/validation/): "This method should return the cleaned value obtained from cleaned_data..."

kumar303:
it would be great if you could add a live server test that hits the sandbox braintree API and generates a real error. This would provide a nice smoke test against all the ``errors.deep_errors...`` private property access that we do.

lib/buyers/forms.py
```diff
@@ -12,10 +12,10 @@ def clean_pin(pin):
         return pin
 
     if not len(pin) == 4:
-        raise forms.ValidationError(PIN_4_NUMBERS_LONG)
+        raise forms.ValidationError(PIN_4_NUMBERS_LONG, code='invalid')
 
     if not pin.isdigit():
-        raise forms.ValidationError(PIN_ONLY_NUMBERS)
+        raise forms.ValidationError(PIN_ONLY_NUMBERS, code='invalid')
```
kumar303: could this code be more specific like ``invalid_pin`` ?

lib/buyers/serializers.py
```diff
@@ -44,7 +44,8 @@ def validate_uuid(self, attrs, source):
             qs = qs.exclude(pk=self.object.pk)
 
         if qs.exists():
-            raise serializers.ValidationError(BUYER_UUID_ALREADY_EXISTS)
+            raise serializers.ValidationError(BUYER_UUID_ALREADY_EXISTS,
+                                              code='not_unique')
```
kumar303: ``buyer_not_unique`` would be less ambiguous

solitude/exceptions.py
```diff
@@ -1,34 +1,29 @@
...
-        }
-        return Response(res, status=400)
+    if hasattr(exc, 'formatter'):
+        try:
+            return Response(exc.formatter(exc).format(),
+                            status=getattr(exc, 'status_code', 422))
+        except:
+            # If the formatter fails, fall back to the standard
+            # error formatting.
+            log.error('Failed to use formatter.')
```
kumar303: if you made this ``log.exception`` then it would include the traceback which I think could be helpful here.

kumar303:
This is cool. I like the namespacing and the ability to parse the new format in the client by checking the status code.

r+wc

solitude/exceptions.py
andymckay: Thats what I was wanting and thought error did, thanks.

lib/buyers/forms.py
andymckay: We could, but I think that diverges from the Django standard where by the code is a generic term not specific to the field:

https://github.com/django/django/search?utf8=%E2%9C%93&q=default_error_messages&type=Code

You've got the code and the field key to pull things back together. If we go for this standard, it could be possible that people would look at the required field and assume it should be like that. Then I have start overriding required, as pin_required.

lib/brains/forms.py
andymckay: good catch

lib/buyers/forms.py
kumar303: oh I see. I guess it makes sense in the context of which field triggered the error. I was thinking it would be nice to use codes like this somewhere in the UI (recessed into the background) to compliment user errors so that QA reports would be useful at a glance. This approach has been really helpful for webpay.

andymckay:
There's no point in adding in a live test because once we move to Braintree using the proxy in #414, they won't work any more, because to ping Braintree you need both solitude and solitude-auth. This is a test for when we are able to send up multiple docker containers at the same time.

PR #419, opened 2015-05-16
copy the git rev into the container

andymckay:


kumar303:
that's pretty sweet!

r+

kumar303:
Hmm, will this though on source only containers? I guess git needs to be installed and the ``.git`` part of the source needs to be in the container.

andymckay:
If the data container has git, it should work.

PR #422, opened 2015-05-20
add in a webhook handler

andymckay:
fixes mozilla/payments-service#16

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -228,7 +228,6 @@ Some information about the subscripton is stored in solitude.
 
 .. http:get:: /braintree/mozilla/subscription/<subscription id>/
 
-
     .. code-block:: json
 
         {
@@ -261,3 +260,25 @@ Some information about the subscripton is stored in solitude.
     :query paymethod__braintree_buyer: the primary key of the braintree buyer.
     :query paymethod__braintree_buyer__buyer: the primary key of the buyer.
     :query seller_product: the primary key of the product.
+
+Webhook
+-------
+
+When Braintree completes certain actions, they will make a request to the
+configured webhook URL. That will be `<payments-service https://github.com/mozilla/payments-service/>`_
+which then passes it on to this endpoint. For more information see the
+`Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks>`_.
+
+.. http:get:: /braintree/webhook/
+
+    :query bt_challenge string: the bt_challenge issued by Braintree.
+    :>json string: a token returned by the Braintree verify API.
+
+    :status 200: token verified and returned.
+
+.. http:post:: /braintree/webhook/
+
+    :<json bt_signature: the bt_signature issued by Braintree.
+    :<json bt_payload: the bt_payload issued by Braintree.
+
+    :status 204: webhook parsed successfully.
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -90,3 +90,23 @@ class SubscriptionForm(forms.Form):
                 'url': 'mozilla.org'
             }
         }
+
+
+class WebhookVerifyForm(forms.Form):
+    bt_challenge = forms.CharField()
+
+    @property
+    def braintree_data(self):
+        return self.cleaned_data['bt_challenge']
+
+
+class WebhookParseForm(forms.Form):
+    bt_signature = forms.CharField()
+    bt_payload = forms.CharField()
+
+    @property
+    def braintree_data(self):
+        return (
+            self.cleaned_data['bt_signature'],
+            self.cleaned_data['bt_payload']
+        )
--- a/lib/brains/tests/test_webhook.py
+++ b/lib/brains/tests/test_webhook.py
@@ -1,0 +1,44 @@
+from django.core.urlresolvers import reverse
+
+from braintree.webhook_notification import WebhookNotification
+from braintree.webhook_notification_gateway import WebhookNotificationGateway
+from nose.tools import eq_
+
+from lib.brains.tests.base import BraintreeTest
+
+
+def parsed(**kwargs):
+    data = {'kind': 'test', 'subject': ''}
+    data.update(kwargs)
+    return WebhookNotification(None, data)
+
+
+def example(**kwargs):
+    data = {'bt_signature': 'signature', 'bt_payload': 'payload'}
+    data.update(kwargs)
+    return data
+
+
+class TestWebhook(BraintreeTest):
+    gateways = {'notify': WebhookNotificationGateway}
+
+    def setUp(self):
+        super(TestWebhook, self).setUp()
+        self.url = reverse('braintree:webhook')
+
+    def test_allowed(self):
+        self.allowed_verbs(self.url, ['get', 'post'])
+
+    def test_get_missing(self):
+        eq_(self.client.get(self.url).status_code, 422)
+
+    def test_get_ok(self):
+        self.mocks['notify'].verify.return_value = 'foo'
+        eq_(self.client.get(self.url + '?bt_challenge=x').json, 'foo')
+
+    def test_post_missing(self):
+        eq_(self.client.post(self.url, data={}).status_code, 422)
+
+    def test_post_ok(self):
+        self.mocks['notify'].parse.return_value = parsed()
+        eq_(self.client.post(self.url, data=example()).status_code, 204)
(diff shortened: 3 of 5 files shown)


docs/topics/braintree.rst
```diff
@@ -261,3 +260,27 @@ Some information about the subscripton is stored in solitude.
     :query paymethod__braintree_buyer: the primary key of the braintree buyer.
     :query paymethod__braintree_buyer__buyer: the primary key of the buyer.
     :query seller_product: the primary key of the product.
+
+Webhook
+-------
+
+When Braintree completes certain actions, webhooks will be sent. For more information
```
kumar303: maybe more specific like: "When Braintree completes...they will post data to our configured webhook URL"

docs/topics/braintree.rst
```diff
@@ -261,3 +260,27 @@ Some information about the subscripton is stored in solitude.
...
+see the `Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks>`_.
+
+.. http:get:: /braintree/webhook/verify/
+
+    .. code-block:: json
+
+        "some-token"
+
+    :<json bt_challenge: the bt_challenge issued by Braintree.
```
kumar303: I can't find documentation on this in the link above. Do they really send JSON data in a GET? Or is it a query string parameter?

docs/topics/braintree.rst
```diff
@@ -261,3 +260,27 @@ Some information about the subscripton is stored in solitude.
...
+    :>json string: a token returned by the Braintree verify API.
+
+    :status 200: token verified and returned.
+
+.. http:post:: /braintree/webhook/parse/
+
+    :<json bt_signature: the bt_signature issued by Braintree.
```
kumar303: docs say these are url-encoded POST params, not JSON

docs/topics/braintree.rst
andymckay: It is a GET, but bear in mind that the request comes to payments service which then sends the request on solitude.

Reading the payments service PR, it is indeed sending a query param, will change.

https://github.com/mozilla/payments-service/pull/21/files#diff-340a6a494441afe57ca1271942b853c3R112

docs/topics/braintree.rst
andymckay: Which doc, the Braintree one? Again goes through payment-service which posts using curling.

https://github.com/mozilla/payments-service/pull/21/files#diff-340a6a494441afe57ca1271942b853c3R116

docs/topics/braintree.rst
andymckay: Sure, and something like "posts to the Webhook URL which is payments-service, which then forwards on to solitude."

lib/brains/forms.py
```diff
@@ -85,3 +85,23 @@ def braintree_data(self):
...
+
+
+class WebhookParseForm(forms.Form):
+    bt_signature = forms.CharField()
+    bt_payload = forms.CharField()
+
+    @property
+    def braintree_data(self):
+        return (
+            self.cleaned_data['bt_signature'],
```
kumar303: this will be a Unicode object but the sample code in the docs show encoding this as ascii before passing it to the API for verification. Does it work without doing that?

docs/topics/braintree.rst
kumar303: ah, right, I see now

docs/topics/braintree.rst
kumar303: yeah, my bad. This is correct.

lib/brains/tests/test_webhook.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_allowed(self):
+        self.allowed_verbs(self.url, ['get'])
+
+    def test_missing(self):
+        eq_(self.client.get(self.url).status_code, 422)
+
+    def test_ok(self):
+        self.mocks['notify'].verify.return_value = 'foo'
+        eq_(self.client.get(self.url + '?bt_challenge=x').json, 'foo')
```
kumar303: in cases like this I think it's helpful to add an assertion that the correct call was made, like ``self.mocks['notify'].verify.assert_called_with('x')``.

kumar303: should this client request pass in a query string parameter or JSON?

lib/brains/tests/test_webhook.py
```diff
@@ -0,0 +1,55 @@
...
+from lib.brains.tests.base import BraintreeTest
+
+
+def parsed(**kwargs):
+    data = {'kind': 'test', 'subject': ''}
+    data.update(kwargs)
+    return WebhookNotification(None, data)
+
+
+def example(**kwargs):
```
kumar303: ``example_notice`` would make more sense to me

lib/brains/tests/test_webhook.py
```diff
@@ -0,0 +1,55 @@
+from django.core.urlresolvers import reverse
+
+from braintree.webhook_notification import WebhookNotification
+from braintree.webhook_notification_gateway import WebhookNotificationGateway
+from nose.tools import eq_
+
+from lib.brains.tests.base import BraintreeTest
+
+
+def parsed(**kwargs):
```
kumar303: ``parsed_notice`` would make more sense to me. I was confused when first reading this code.

lib/brains/views/webhook.py
```diff
@@ -0,0 +1,31 @@
...
+log = getLogger('s.brains')
+
+
+@api_view(['POST'])
+def parse(request):
+    form = WebhookParseForm(request.DATA)
+    if not form.is_valid():
+        raise FormError(form.errors)
+
+    parsed = get_client().WebhookNotification.parse(*form.braintree_data)
```
kumar303: I'm not a fan of putting the braintree call data into the form because it seems like an unnecessary abstraction. You can't glance at this call and see what data is passed in, you have to switch to the form code. It's fine to keep it as is because it's consistent with the rest of solitude, just thought I'd mention it.

lib/brains/views/webhook.py
```diff
@@ -0,0 +1,31 @@
...
+    log.info('Received webhook: {p.kind}.'.format(p=parsed))
```
kumar303: it would be nice to log the complete parsed notice. Maybe as a separate debug call?

lib/brains/views/webhook.py
```diff
@@ -0,0 +1,31 @@
...
+    return Response(status=204)
+
+
+@api_view(['GET'])
+def verify(request):
+    form = WebhookVerifyForm(request.QUERY_PARAMS)
+    if not form.is_valid():
+        raise FormError(form.errors)
+
+    res = get_client().WebhookNotification.verify(form.braintree_data)
```
kumar303: Since we will have no visibility into this response (only Braintree does), I think you should log it.

kumar303:
r+wc

docs/topics/braintree.rst
```diff
@@ -261,3 +260,27 @@ Some information about the subscripton is stored in solitude.
     :query paymethod__braintree_buyer: the primary key of the braintree buyer.
     :query paymethod__braintree_buyer__buyer: the primary key of the buyer.
     :query seller_product: the primary key of the product.
+
```
kumar303: hmm, I wonder if you should add a rST comment here mentioning that the ``payments-service`` docs will link to this anchor per https://github.com/mozilla/payments-service/pull/21/files#diff-7d64b15bbaac9444c9da295924bf42bfR188

lib/brains/forms.py
andymckay: Yep, using ngrok, was able to test this end to end.

andymckay:
```
solitude:  s.brains:INFO payments-service:- Received verification response: r7ms2hhp9pnz8srd|ac686e083577fda612d4f9b1fec1c764d775c0b1 :/srv/solitude/lib/brains/views/webhook.py:37
[22/May/2015 15:21:54]"GET /braintree/webhook/?bt_challenge=f9c935c311660107b8a1 HTTP/1.1" 200 59
solitude:  s.brains:INFO payments-service:- Received webhook: subscription_went_active. :/srv/solitude/lib/brains/views/webhook.py:25
[22/May/2015 15:23:01]"POST /braintree/webhook/ HTTP/1.1" 204 0
solitude:  s.brains:INFO payments-service:- Received webhook: subscription_charged_successfully. :/srv/solitude/lib/brains/views/webhook.py:25
[22/May/2015 15:23:01]"POST /braintree/webhook/ HTTP/1.1" 204 0
```

docs/topics/braintree.rst
andymckay: Could I do that through a label?

PR #424, opened 2015-05-20
move name to descriptor

### LINKED ISSUE #423: Fix braintree product descriptors
A [descriptor](https://developers.braintreepayments.com/javascript+python/reference/request/subscription/create#descriptor) is what shows up on a buyer's bill -- the more info it has the less likely they are to dispute the charge. Currently the descriptor is causing an error. Here is a basic descriptor that passes without error but ``product`` needs to be replaced with an actual product name.

````diff
diff --git a/lib/brains/forms.py b/lib/brains/forms.py
index 645af52..784957d 100644
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -78,5 +78,5 @@ class SubscriptionForm(forms.Form):
             'payment_method_token': self.cleaned_data['paymethod'].provider_id,
             'plan_id': self.seller_product.external_id,
             'trial_period': False,
-            'name': 'Mozilla',
+            'descriptor': {'name': 'Mozilla*product', 'url': 'mozilla.org'},
         }

````

### PR DESCRIPTION
**Pull request by andymckay**
fixes #423

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -83,5 +83,10 @@ class SubscriptionForm(forms.Form):
             'payment_method_token': self.cleaned_data['paymethod'].provider_id,
             'plan_id': self.seller_product.external_id,
             'trial_period': False,
-            'name': 'Mozilla',
+            'descriptor': {
+                # TODO: figure out how to get product in here
+                # https://github.com/mozilla/payments/issues/57
+                'name': 'Mozilla*product',
+                'url': 'mozilla.org'
+            }
         }
--- a/lib/brains/tests/test_subscription.py
+++ b/lib/brains/tests/test_subscription.py
@@ -54,7 +54,10 @@ class TestSubscriptionMethod(BraintreeTest):
         data = {
             'payment_method_token': mock.ANY,
             'plan_id': 'brick',
-            'name': 'Mozilla',
+            'descriptor': {
+                'name': 'Mozilla*product',
+                'url': 'mozilla.org'
+            },
             'trial_period': False,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/brains/forms.py [hunk 1]
```diff
@@ -83,5 +83,10 @@ def braintree_data(self):
             'payment_method_token': self.cleaned_data['paymethod'].provider_id,
             'plan_id': self.seller_product.external_id,
             'trial_period': False,
-            'name': 'Mozilla',
+            'descriptor': {
+                # TODO: figure out how to get product in here
+                # https://github.com/mozilla/payments/issues/57
+                'name': 'Mozilla *product',
```
kumar303: I tested this and oddly it cannot have a space in it. It must be ``Mozilla*product`` where the final string is less than 22 characters. Maybe until we fix #57 we could make it ``Mozilla*subscription`` but whatever, just a placeholder.

andymckay: I'll not that the given example has a space in: https://developers.braintreepayments.com/javascript+python/reference/request/subscription/create#dynamic-descriptors

lib/brains/tests/test_subscription.py
```diff
@@ -54,7 +54,10 @@ def test_ok(self):
         data = {
             'payment_method_token': mock.ANY,
             'plan_id': 'brick',
-            'name': 'Mozilla',
+            'descriptor': {
+                'name': 'Mozilla *product',
```
kumar303: test need fixing

kumar303:
r+wc

PR #426, opened 2015-05-21
looks like the externalCPTransID changed

andymckay:
Example from today:

```
May 21 20:49:52 localhost6.localdomain: [10.8.83.211] w.bango:DEBUG Bango notice: '\xef\xbb\xbf<?xml version="1.0" encoding="utf-8"?>...<item name="externalCPTransID" value="webpay:...
```


robhudson:
r+

PR #427, opened 2015-05-22
bump django-extensions

andymckay:
* the old version of django-extensions was causing a failure in schematic migrations, which was breaking solitude start up


.travis.yml
```diff
@@ -7,7 +7,7 @@ before_install:
 install:
 - DEB_HOST_MULTIARCH=x86_64-linux-gnu pip install -I --exists-action=w "git+git://anonscm.debian.org/collab-maint/m2crypto.git@debian/0.21.1-3#egg=M2Crypto"
 - pip install --no-deps MySQL-Python==1.2.5 --use-mirrors
```
kumar303: I think you also need to install peep here

kumar303:
r+wc

.travis.yml
andymckay: oh yeah, doh

PR #433, opened 2015-05-26
Convert to travis container build for speed

kumar303:


kumar303:
Oh ha, my bad. I didn't realize that container builds cannot use sudo at all. That's weird.

PR #434, opened 2015-05-26
Respond with credit card errors (fixes #431)

### LINKED ISSUE #431: Handle credit card processing errors
We can handle [validation errors](https://developers.braintreepayments.com/javascript+python/reference/general/validation-errors/overview) ok but not [credit card processing errors](https://developers.braintreepayments.com/javascript+python/reference/general/processor-responses/authorization-responses).

To reproduce:

Enter a [test card](https://developers.braintreepayments.com/javascript+python/reference/general/testing) to produce a processor error such as 4000111111111115 which results in processor declined.

Solitude currently responds with: ``{..., "braintree": {}}`` because ``self.error.result.errors.deep_errors`` is an empty list. We need to parse ``self.error.result.credit_card_verification`` to get the processor code so it can be returned to the UI for display and localization.

kumar303:


lib/brains/errors.py
```diff
@@ -21,6 +24,16 @@ def format(self):
                 'message': error.message
             })
 
+        cc_result = self.error.result.credit_card_verification
+        if cc_result:
+            log.debug('credit card processing error: {r}'.format(r=cc_result))
+            errors['__all__'].append({
```
andymckay: fyi: `from django.core.exceptions import NON_FIELD_ERRORS` is `__all__`

lib/brains/errors.py
```diff
@@ -21,6 +24,16 @@ def format(self):
...
+                # Prefix the actual error code with `cc-' so that it's sort
+                # of namespaced against other types of error codes.
```
andymckay: Thats a good idea, should probably add a comment to the docs on errors about that?

andymckay:
r+ thanks

lib/brains/errors.py
kumar303: Nifty. That would be funny if it ever changes.

lib/brains/errors.py
kumar303: It feels dirty. I was thinking that the UI might need something like that for localization though. I'll take a look at the docs.

PR #441, opened 2015-05-28
process a subscription charged event

### LINKED ISSUE #412: Web hook handler for subscription charged
Create a web hook handler for subscription charged successfully.

andymckay:
fixes #412


andymckay:
This is going to change.

Issue #444, opened 2015-06-01
Add in Bango into solitude-auth


andymckay-limited-access:
Make sure that Bango can use the solitude-auth so that we can remove the proxy.

PR #448, opened 2015-06-01
Add bango

andymckay:
for mozilla/solitude#444


andymckay:
all this does is move the HTTP value down into the settings to allow re-use by braintree

kumar303:
r+

Issue #449, opened 2015-06-02
Logging of all webhooks


andymckay-limited-access:
I don't think we can log all the web hooks all the time because they are very verbose, but for the next few weeks, it would really help if I could see all the possible web hooks.

So this is a flag in solitude to write out a detailed log somewhere. Then we can see if we can trigger some web hooks.

PR #450, opened 2015-06-02
more logging

andymckay:
for #449

### DIFF
--- a/lib/brains/views/webhook.py
+++ b/lib/brains/views/webhook.py
@@ -1,12 +1,15 @@
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
 
+from django.conf import settings
+
 from lib.brains.client import get_client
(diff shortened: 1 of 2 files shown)


lib/brains/views/webhook.py
```diff
@@ -23,6 +25,8 @@ def parse(request):
 
     parsed = get_client().WebhookNotification.parse(*form.braintree_data)
     log.info('Received webhook: {p.kind}.'.format(p=parsed))
+    if settings.BRAINTREE_WEBHOOK_LOGGING:
+        log.info(parsed)
```
kumar303: couldn't this just be ``log.debug(...)`` ?

kumar303:
I think debug logging would be fine instead.

r+wc

kumar303:
I personally like to leave logs in debug mode for development so a setting might be nice to turn this off since it seems like it will be very verbose.

PR #451, opened 2015-06-02
process a subscription charged event

andymckay:
this is the newly modeled event based on mozilla/payments#28 which does not require a transaction to be made in payments-service

### DIFF
--- a/lib/brains/tests/test_webhook.py
+++ b/lib/brains/tests/test_webhook.py
@@ -1,16 +1,55 @@
+<<<<<<< HEAD
 import base64
+=======
+from decimal import Decimal
+>>>>>>> process a subscription charged event
 
 from django.core.urlresolvers import reverse
 from django.test.utils import override_settings
 
 from braintree.webhook_notification import WebhookNotification
+<<<<<<< HEAD
 from nose.tools import eq_
+=======
+from braintree.webhook_notification_gateway import WebhookNotificationGateway
+from mock import patch
+from nose.tools import eq_, ok_
+>>>>>>> process a subscription charged event
 
-from lib.brains.tests.base import BraintreeTest
+from lib.brains.models import BraintreeSubscription
+from lib.brains.tests.base import (
+    BraintreeTest, create_braintree_buyer, create_method, create_seller)
+from lib.brains.webhooks import Processor
+from lib.transactions import constants
+from lib.transactions.models import Transaction
 
 
 def parsed(**kwargs):
-    data = {'kind': 'test', 'subject': ''}
+    data = {'kind': 'test', 'subject': {}}
+    data.update(kwargs)
+    return WebhookNotification(None, data)
+
+
+subscription = {
+    'subscription': {
+        'id': 'some-bt:id',
+        'transactions': [
+            {
+                'id': 'bt:id',
+                'amount': 10,
+                'tax_amount': 1,
+                'currency_iso_code': 'USD'
+            }
+        ]
+    }
+}
+
+
+def parsed_subscription(**kwargs):
+    data = {
+        'kind': 'subscription_charged_successfully',
+        'subject': subscription
+    }
     data.update(kwargs)
     return WebhookNotification(None, data)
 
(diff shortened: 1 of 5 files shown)


lib/brains/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+        # Its currently unclear if the most recent transaction is first or
+        # last in the last of transactions. Will need to see how multiple
+        # transactions look.
+        their_transaction = their_subscription.transactions[0]
+        transaction = Transaction.objects.create(
+            amount=their_transaction.amount,
+            buyer=subscription.paymethod.braintree_buyer.buyer,
+            provider=constants.PROVIDER_BRAINTREE,
+            seller=subscription.seller_product.seller,
+            status=constants.STATUS_CHECKED,
```
andymckay: This is one area that I need to investigate, I don't think we can automatically assume that the transaction is correct. This will work for now.

lib/brains/tests/test_webhook.py
```diff
@@ -39,6 +66,42 @@ def test_get_ok(self):
...
+        Processor(parsed(kind=self.kind, subject=self.sub)).process()
+        eq_(Transaction.objects.get().status, constants.STATUS_CHECKED)
+
+    def test_sets_amount(self):
+        Processor(parsed(kind=self.kind, subject=self.sub)).process()
+        eq_(Transaction.objects.get().amount, Decimal('10'))
+
+    def test_sets_uid_support(self):
+        Processor(parsed(kind=self.kind, subject=self.sub)).process()
+        eq_(Transaction.objects.get().uid_support, 'bt:id')
```
kumar303: where is ``bt:id`` set? Maybe reference it with a variable to make the tests less brittle?

lib/brains/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+        We have to:
+        * find the subscription and create a transaction for it.
+        """
+        their_subscription = self.webhook.subscription
+        try:
+            subscription = BraintreeSubscription.objects.get(
+                provider_id=their_subscription.id
+            )
+        except ObjectDoesNotExist:
+            log.exception('No transaction found: {}'
```
kumar303: :fried_shrimp:

kumar303: oh wait! I didn't know empty brackets was a thing in Python. Heh, nevermind.

kumar303: This error message should say 'No subscription...' right?

lib/brains/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+                          .format(their_subscription.id))
+            raise
+
+        # Its currently unclear if the most recent transaction is first or
+        # last in the last of transactions. Will need to see how multiple
+        # transactions look.
+        their_transaction = their_subscription.transactions[0]
```
kumar303: maybe add a TODO and/or issue for this? Wouldn't we also want to loop through all transactions?

lib/brains/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+        transaction = Transaction.objects.create(
+            amount=their_transaction.amount,
```
kumar303: do we not get currency as well?

lib/brains/webhooks.py
```diff
@@ -0,0 +1,66 @@
...
+            buyer=subscription.paymethod.braintree_buyer.buyer,
+            provider=constants.PROVIDER_BRAINTREE,
+            seller=subscription.seller_product.seller,
+            status=constants.STATUS_CHECKED,
+            type=constants.TYPE_PAYMENT,
+            uid_support=their_transaction.id
```
kumar303: I don't see ``seller_product``. We need to add that. It should be tied to the plan ID in this case, I think.

kumar303: should this be ``uid_pay`` instead? I forget. The column names are confusing.

kumar303:
r- for missing transaction data (unless product is optional? doesn't seem like it could be)

lib/brains/tests/test_webhook.py
andymckay: its just a string set up above in the same file, don't normally make vars for everything

lib/brains/webhooks.py
andymckay: yup, thanks

lib/brains/webhooks.py
andymckay: We do, I was hoping to just get this in as the first pass and work my way through all the hooks in later commits.

lib/brains/webhooks.py
andymckay: we do, will add

lib/brains/webhooks.py
andymckay: `uid_pay` is for the pre transaction key in Bango, the paykey in Paypal... there isn't a parallel in braintree the more I though about it. I think `uid_support` is the right place, since its the id we'll use to link up with Braintree.

lib/brains/webhooks.py
andymckay: So on the first flow we have right now you get a subscription_charged event and one transaction and that's it, so this works for basic flow.

kumar303:
r+, transaction looks right

PR #453, opened 2015-06-04
pass verification onto solitude-auth

### LINKED ISSUE #452: Webhook failing
I think when the change to using solitude-auth has caused:

```
  File "/srv/solitude/lib/brains/views/webhook.py", line 25, in parse
    parsed = get_client().WebhookNotification.parse(*form.braintree_data)
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/braintree/webhook_notification.py", line 35, in parse
    return Configuration.gateway().webhook_notification.parse(signature, payload)
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/braintree/webhook_notification_gateway.py", line 24, in parse
    self.__validate_signature(signature, payload)
  File "/opt/rh/python27/root/usr/lib/python2.7/site-packages/braintree/webhook_notification_gateway.py", line 42, in __validate_si
gnature
    raise InvalidSignatureError("no matching public key")
InvalidSignatureError: no matching public key
```

andymckay:
fixes #452 along with mozilla/solitude-auth#4

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -1,10 +1,16 @@
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+
+import requests
 
 from lib.brains.models import BraintreeBuyer, BraintreePaymentMethod
 from lib.buyers.models import Buyer
 from lib.sellers.models import SellerProduct
+from solitude.base import getLogger
 from solitude.related_fields import PathRelatedFormField
+
+log = getLogger('s.brains')
 
 
 class BuyerForm(forms.Form):
@@ -99,6 +105,19 @@ class WebhookVerifyForm(forms.Form):
     def braintree_data(self):
         return self.cleaned_data['bt_challenge']
 
+    def clean_bt_challenge(self):
+        res = requests.get(
+            settings.BRAINTREE_PROXY + '/verify',
+            params={'bt_challenge': self.cleaned_data['bt_challenge']}
+        )
+        if res.status_code != 200:
+            log.error('Did not receive a 204 from solitude-auth, got: {}'
+                      .format(res.status_code))
+            raise forms.ValidationError(
+                'Did not pass verification', code='invalid')
+        self.response = res.content
+        return self.cleaned_data['bt_challenge']
+
 
 class WebhookParseForm(forms.Form):
     bt_signature = forms.CharField()
@@ -110,3 +129,12 @@ class WebhookParseForm(forms.Form):
             self.cleaned_data['bt_signature'],
             self.cleaned_data['bt_payload']
         )
+
+    def clean(self):
+        res = requests.post(settings.BRAINTREE_PROXY + '/parse', self.data)
+        if res.status_code != 204:
+            log.error('Did not receive a 204 from solitude-auth, got: {}'
+                      .format(res.status_code))
+            raise forms.ValidationError(
+                'Did not pass verification', code='invalid')
+        return self.cleaned_data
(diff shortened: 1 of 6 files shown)


lib/brains/forms.py
```diff
@@ -99,6 +105,18 @@ class WebhookVerifyForm(forms.Form):
...
 
+    def clean_bt_challenge(self):
+        res = requests.get(
+            settings.BRAINTREE_PROXY + '/verify',
+            params={'bt_challenge': self.data['bt_challenge']}
+        )
+        if res.status_code != 200:
+            log.error('Did not receive a 204 from solitude-auth, got: {}'
+                      .format(res.status_code))
+            raise forms.ValidationError('Did not pass verification')
```
kumar303: are we still adding ``code=...`` ? I forget

lib/brains/forms.py
```diff
@@ -99,6 +105,18 @@ class WebhookVerifyForm(forms.Form):
     def braintree_data(self):
         return self.cleaned_data['bt_challenge']
 
+    def clean_bt_challenge(self):
+        res = requests.get(
+            settings.BRAINTREE_PROXY + '/verify',
+            params={'bt_challenge': self.data['bt_challenge']}
```
kumar303: do you have to use ``self.data`` so that it's binary?

lib/brains/tests/base.py
```diff
@@ -126,6 +128,28 @@ def cleanUpBrains(self):
...
 
+    def cleanUpRequest(self):
+        self.req.patcher.stop()
+
+    def patch_webhook_forms(self):
+        request = mock.patch('lib.brains.forms.requests', name='test.forms')
+        self.req = request.start()
+        self.req.exceptions = requests.exceptions
+        self.req.patcher = request
+        self.addCleanup(self.cleanUpRequest)
```
kumar303: could also be ``self.addCleanup(self.req.patcher.stop)``

lib/brains/forms.py
andymckay: I should do, good call.

lib/brains/forms.py
andymckay: I couldn't remember if its in cleaned_data when its being cleaned_, probably is right?

lib/brains/tests/base.py
andymckay: Yeah I'm hoping to add more to cleanUpRequest though.

lib/brains/forms.py
```diff
@@ -110,3 +128,11 @@ def braintree_data(self):
             self.cleaned_data['bt_signature'],
             self.cleaned_data['bt_payload']
         )
+
+    def clean(self):
+        res = requests.post(settings.BRAINTREE_PROXY + '/parse', self.data)
```
kumar303: this was meant to be ``self.data['bt_payload']``, no?

kumar303: Oh, nevermind, I see that you are just passing through the signature and payload

lib/brains/forms.py
```diff
@@ -110,3 +128,11 @@ def braintree_data(self):
...
+        if res.status_code != 204:
+            log.error('Did not receive a 204 from solitude-auth, got: {}'
+                      .format(res.status_code))
+            raise forms.ValidationError('Did not pass verification')
```
kumar303: Since this is the point at which we can trust the payload, it might be more convenient to unpack it here:

    data = base64.decodestring(self.cleaned_data['bt_payload'])
    data = XmlUtil.dict_from_xml(data)
    self.cleaned_data['bt_payload'] = data

kumar303: but you'd have to protect against empty input, I suppose (which will cause a decode error before raising a 'field is required' error, maybe)

kumar303:
r+wc

Issue #456, opened 2015-06-04
Use public_id to map Braintree subscription plans more uniquely


kumar303:
The braintree configuration script puts the unique ID for a product (i.e. a Braintree subscription plan) into ``external_id`` but this is not unique. It should put them in ``public_id`` which is unique. That way we can look them up easier. 

Things to do:
* config script needs an update
* the subscription form lookup needs adjusting
* everyone needs to dump and re-run the config

PR #457, opened 2015-06-04
flip external_id and public_id around

andymckay:
* also throw an error if the BRAINTREE_PROXY isn't defined

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -75,7 +75,7 @@ class SubscriptionForm(forms.Form):
         data = self.cleaned_data['plan']
 
         try:
-            obj = SellerProduct.objects.get(external_id=data)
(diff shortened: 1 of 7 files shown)


andymckay:
fixes #456

lib/brains/forms.py
```diff
@@ -75,7 +75,7 @@ def clean_plan(self):
         data = self.cleaned_data['plan']
 
         try:
-            obj = SellerProduct.objects.get(external_id=data)
+            obj = SellerProduct.objects.get(public_id=data)
         except ObjectDoesNotExist:
```
kumar303: I just noticed that this could use a logging statement to show the plan ID used to do the lookup when it fails. Otherwise, debugging might be difficult.

kumar303:
r+wc, thanks

PR #458, opened 2015-06-05
make sure we start at the beginning of the day

andymckay:


kumar303:
this might be a little more straightforward by putting this in setUp:

    self.first = Transaction.objects.create(...)
    self.start_date = self.first.created.date()

but not completely necessary

kumar303:
r+wc

andymckay:
I like that because it won't fail at 11:59.

kumar303:
Yep, I was going to mention that but no one should be running tests at 11:59 ;)

PR #465, opened 2015-06-06
look at all the transactions on a webhook

### LINKED ISSUE #429: Cope with repeated transactions to a subscription
Braintree will send a notification each month when a subscription is charged. In this case a transaction won't exist in our table because we didn't create it all in one flow. We'll need to cope with that.

andymckay:
* instead of looking at just the first one, look at them all
* ignore repeats
* save others with some info
* fixes #429

### DIFF
--- a/lib/brains/tests/test_webhook.py
+++ b/lib/brains/tests/test_webhook.py
@@ -16,34 +16,34 @@ from lib.transactions import constants
 from lib.transactions.models import Transaction
 
 
-def parsed(**kwargs):
-    data = {'kind': 'test', 'subject': {}}
+def notification(**kwargs):
+    data = {
+        'kind': 'test',
+        'subject': {}
+    }
     data.update(kwargs)
     return WebhookNotification(None, data)
 
 
-subscription = {
-    'subscription': {
-        'id': 'some-bt:id',
-        'transactions': [
-            {
-                'id': 'bt:id',
-                'amount': 10,
-                'tax_amount': 1,
-                'currency_iso_code': 'USD'
-            }
-        ]
-    }
-}
-
-
-def notification(**kwargs):
+def transaction(**kwargs):
     data = {
-        'kind': 'subscription_charged_successfully',
-        'subject': subscription
+        'id': 'bt:id',
+        'amount': 10,
+        'tax_amount': 1,
+        'currency_iso_code': 'USD',
+        'status': 'settled'
     }
     data.update(kwargs)
     return data
+
+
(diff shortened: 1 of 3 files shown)


lib/brains/tests/test_webhook.py
```diff
@@ -109,14 +109,83 @@ def setUp(self):
...
+        eq_(trans.status, constants.STATUS_FAILED)
+        eq_(trans.status_reason, 'gateway_rejected wat')
+
+    def test_repeated(self):
+        sub = subscription(transactions=[transaction(status='settled')])
+        self.process(sub)
+        self.process(sub)
+        eq_(Transaction.objects.count(), 1)
+
+    def test_repeated_changed(self):
```
kumar303: Since there aren't any comments and this test name is vague, it took me a while to figure out what it's testing. Maybe: ``test_trans_status_must_not_change()`` ?

lib/brains/webhooks.py
```diff
@@ -50,19 +59,82 @@ def process_subscription_charged_successfully(self):
...
+        * submitted_for_settlement
+        * voided
+
+        All of the ignored states are temporary statuses and we don't
+        really care about those.
+        """
+        for their_transaction in their_subscription.transactions:
+            status = their_transaction.status
+            if status not in settings.BRAINTREE_TRANSACTION_STATUSES:
+                log.info('Ignoring transaction status: {}'.format(status))
```
kumar303: could you also log the subscription ID or something to tie it to?

lib/brains/webhooks.py
```diff
@@ -50,19 +59,82 @@ def process_subscription_charged_successfully(self):
...
+
+            try:
+                our_transaction = Transaction.objects.get(
+                    uid_support=their_transaction.id,
+                )
+                log.info('Transaction exists: {}'.format(our_transaction.pk))
+                # Just a maybe pointless sanity check that the status they are
+                # sending in their transaction matches our record.
+                if our_transaction.status != our_status:
+                    raise ValueError('Status: {} does not match: {}'
```
kumar303: could you include the subscription or transaction ID in this message?

kumar303:
r+wc

lib/brains/tests/test_webhook.py
andymckay: sure

lib/brains/webhooks.py
andymckay: I'll put that in a seperate log above.

PR #470, opened 2015-06-08
allow lookup by provider_id

### LINKED ISSUE #466: Filter braintree subscriptions by plan ID
We need to look up whether a user is already subscribed to a plan (https://github.com/mozilla/payments-service/pull/48) so it would save some look-ups if we could filter [all subscriptions](https://solitude.readthedocs.org/en/latest/topics/braintree.html#get--braintree-mozilla-subscription-) by braintree plan ID.

andymckay:
fixes #466

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -256,9 +256,11 @@ Some information about the subscripton is stored in solitude.
 
 .. http:get:: /braintree/mozilla/subscription/
 
+    :query active: if the subscription is active.
     :query paymethod: the primary key of the payment method.
     :query paymethod__braintree_buyer: the primary key of the braintree buyer.
     :query paymethod__braintree_buyer__buyer: the primary key of the buyer.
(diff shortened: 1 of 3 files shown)


docs/topics/braintree.rst
```diff
@@ -256,9 +256,11 @@ Some information about the subscripton is stored in solitude.
 
 .. http:get:: /braintree/mozilla/subscription/
 
+    :query active: if the subscription is active.
     :query paymethod: the primary key of the payment method.
     :query paymethod__braintree_buyer: the primary key of the braintree buyer.
     :query paymethod__braintree_buyer__buyer: the primary key of the buyer.
+    :query provider_id: the providers id for this subscription.
```
kumar303: typo: provider's

kumar303: It might be good to say explicitly that this will be the braintree plan ID because that's pretty much it

kumar303:
r+wc

PR #471, opened 2015-06-09
Add braintree cleanup script for development

kumar303:


docs/topics/braintree.rst
```diff
@@ -282,3 +282,10 @@ which then passes it on to this endpoint. For more information see the
     :<json bt_payload: the bt_payload issued by Braintree.
 
     :status 204: webhook parsed successfully.
+
+Development Tips
+----------------
+
+When developing on systems that rely on Braintree data in Solitude
+you can do some cleanup with the ``./manage.py braintree_cleanup`` script.
```
andymckay: Can we call it `clear` or `delete` or something? `cleaning` up is a good thing that i try to encourage my wife and kids to do all the time. They never do of course. But I can understand how someone might thing this is a non-destructive command.

andymckay:
r+wc

docs/topics/braintree.rst
kumar303: heh. I can switch it to ``reset``

PR #472, opened 2015-06-10
return something on webhooks

andymckay:
looking for feedback, concerned about the link between transactions and subscription, as mentioned in the comments

### DIFF
--- a/lib/brains/serializers.py
+++ b/lib/brains/serializers.py
@@ -4,6 +4,7 @@ from rest_framework import serializers
 
 from lib.brains.models import (
     BraintreeBuyer, BraintreePaymentMethod, BraintreeSubscription)
+from lib.transactions.serializers import TransactionSerializer
 from solitude.base import BaseSerializer
 from solitude.related_fields import PathRelatedField
 
@@ -25,6 +26,24 @@ class Namespaced(serializers.Serializer):
             'mozilla': self.mozilla.data,
             'braintree': self.braintree.data
         }
+
+
+class LocalReceipt(serializers.Serializer):
+
+    """
+    Serialize enough information to send the user a receipt in one go.
+    """
+
+    def __init__(self, **kw):
+        self.serial = {
+            'paymethod': LocalPayMethod(instance=kw['paymethod']),
+            'subscription': LocalSubscription(instance=kw['subscription']),
+            'transaction': TransactionSerializer(instance=kw['transaction'])
+        }
+
+    @property
+    def data(self):
+        return dict((k, v.data) for k, v in self.serial.items())
 
 
 class LocalPayMethod(BaseSerializer):
@@ -73,7 +92,13 @@ class Braintree(serializers.Serializer):
 
     @property
     def data(self):
-        return dict([k, getattr(self.object, k)] for k in self.fields)
+        res = {}
+        for field in self.fields:
+            obj = self.object
+            for k in field.split('.'):
+                obj = getattr(obj, k)
+            res[k] = obj
+        return res
 
 
 class PayMethod(Braintree):
@@ -86,3 +111,13 @@ class Customer(Braintree):
 
 class Subscription(Braintree):
     fields = ['id', 'created_at', 'updated_at']
+
+
+class Webhook(Braintree):
+    fields = [
+        'kind',
+        'subscription.billing_period_end_date',
+        'subscription.billing_period_start_date',
+        'subscription.next_billing_date',
+        'subscription.next_billing_period_amount',
+    ]
--- a/lib/brains/tests/test_webhook.py
+++ b/lib/brains/tests/test_webhook.py
@@ -1,4 +1,5 @@
 import base64
+from datetime import datetime, timedelta
 from decimal import Decimal
 
 from django.core.urlresolvers import reverse
@@ -41,6 +42,11 @@ def subscription(**kwargs):
     data = {
         'id': 'some-bt:id',
         'transactions': [transaction()],
+        'next_billing_date': datetime.today() + timedelta(days=30),
+        'next_billing_period_amount': 10,
+        'billing_period_end_date': datetime.today() + timedelta(days=29),
+        'billing_period_start_date': datetime.today(),
+        'price': 10
     }
     data.update(kwargs)
     return {'subscription': data}
@@ -63,7 +69,20 @@ def example(**kwargs):
     return data
 
 
-@override_settings(BRAINTREE_PROXY='http://m.o')
+class SubscriptionTest(BraintreeTest):
+
+    def setUp(self):
+        self.buyer, self.braintree_buyer = create_braintree_buyer()
+        self.method = create_method(self.braintree_buyer)
+        self.seller, self.seller_product = create_seller()
+        self.braintree_sub = BraintreeSubscription.objects.create(
+            paymethod=self.method,
+            seller_product=self.seller_product,
+            provider_id='some-bt:id'
+        )
+        self.sub = subscription(id='some-bt:id')
+
+
 class TestWebhook(BraintreeTest):
 
     def setUp(self):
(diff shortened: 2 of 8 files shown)


lib/brains/models.py
```diff
@@ -69,3 +70,12 @@ class Meta(Model.Meta):
...
                        kwargs={'pk': self.pk})
+
+    def get_transactions(self, ids=None):
+        query = (Transaction.objects
+                 .filter(buyer=self.paymethod.braintree_buyer.buyer,
+                         seller_product=self.seller_product))
+        if ids:
+            query = query.filter(id__in=ids)
+
+        return query.order_by('-created')
```
andymckay: I'm not sure this is worth putting on the model, tempted to move it back to webhooks.

lib/brains/webhooks.py
```diff
@@ -30,6 +33,27 @@ def process(self):
...
+        return Namespaced(
+            LocalReceipt(
+                subscription=subscription,
+                paymethod=subscription.paymethod,
+                # TOOD: this does not feel very thread safe.
+                # Transactions on a subscription are:
+                #   Transactions associated with the subscription,
+                #   sorted by creation date with the most recent first.
+                transaction=subscription.get_transactions(
+                    ids=self.transaction_ids).first()
```
andymckay: This should also be filtering by the transaction type, eg: completed transactions. I'm not sure why braintree would send the most recent transaction as a failed one but I don't know.

Also feel like I should do something smart if there's not transaction at all.

lib/brains/webhooks.py
```diff
@@ -19,6 +20,8 @@ class Processor(object):
 
     def __init__(self, webhook):
         self.webhook = webhook
+        self.transaction_ids = []
```
andymckay: I could be storing the transactions themselves instead of using ids and then doing a query. Mind you, I have to sort them by date.

kumar303:
I think we need access to ``number_of_billing_cycles`` from the braintree subscription object so we can say "per month" in the email. I can see subscriptions per year as being a thing.

andymckay:
Will add `number_of_billing_cycles`.

kumar303:
We might also need ``paid_through_date`` from the braintree subscription object to say "Dec 16 - Jan 16 2015" in the email receipt to indicate the paid period.

I'm looking at the email mock up here: http://payments.readthedocs.org/en/latest/design/api/provider.html#payment-method-updated-for-subscription

kumar303:
> We might also need paid_through_date from the braintree subscription object to say "Dec 16 - Jan 16 2015" in the email receipt to indicate the paid period.

Actually, I think we just need ``billing_period_end_date`` and ``billing_period_start_date`` from the [subscription section](https://developers.braintreepayments.com/javascript+python/reference/response/transaction#subscription) of the transaction.

andymckay:
`number_of_billing_cycles` isn't what you need for per month, that's for plans that end in a limited amount of months - eg: 6 month subscription

Not sure how you'll get monthly - payments-config?

kumar303:
ok, yeah, maybe we can just add it to the config

kumar303:
It might be convenient if this API returned the buyer's email address so we don't have to do a look up. I also need to start saving it (see https://github.com/mozilla/payments-service/issues/54)

lib/brains/serializers.py
```diff
@@ -27,6 +28,24 @@ def data(self):
         }
 
 
+class LocalReceipt(serializers.Serializer):
+
+    """
+    Serialize enough information to send the user a receipt in one go.
+    """
+
+    def __init__(self, **kw):
```
kumar303: if this were ``def __init__(self, paymethod, subscription, transaction)`` then you could still pass them in as keywords. It would read a bit cleaner.

lib/brains/webhooks.py
```diff
@@ -30,6 +35,29 @@ def process(self):
...
+        self.processed = True
+
+    @property
+    def data(self):
+        if not self.processed:
+            return
+
+        subscription = self.get_subscription()
+        try:
+            transaction = self.transactions[0]
```
kumar303: This seems a little fragile. At the very least maybe raise an error if there is more than one transaction?

I think I missed something with transactions. I tried to look this up in Braintree's guide to webhooks but couldn't find it. Maybe ping me on this when you are free? I probably need to understand more about how it works.

andymckay:
I triggered a purchase and got this in the webhook (there's an additional print in there):

solitude:  s.brains:INFO payments-service:- Received webhook: subscription_charged_successfully. :/srv/solitude/lib/brains/views/webhook.py:37
solitude:  s.webhooks:INFO payments-service:- Processing event: subscription_charged_successfully :/srv/solitude/lib/brains/webhooks.py:36
solitude:  s.webhooks:INFO payments-service:- Found subscription: 14 :/srv/solitude/lib/brains/webhooks.py:115
solitude:  s.webhooks:INFO payments-service:- Processing transaction status: submitted_for_settlement :/srv/solitude/lib/brains/webhooks.py:161
solitude:  s.webhooks:INFO payments-service:- Transaction created: 17 :/srv/solitude/lib/brains/webhooks.py:208
solitude:  s.webhooks:INFO payments-service:- Found subscription: 14 :/srv/solitude/lib/brains/webhooks.py:115
{'mozilla': {'paymethod': {'resource_pk': 14L, 'resource_uri': u'/braintree/mozilla/paymethod/14/', 'braintree_buyer': u'/braintree/mozilla/buyer/3/', u'id': 14L, 'created': datetime.datetime(2015, 6, 10, 17, 55, 11), 'modified': datetime.datetime(2015, 6, 10, 17, 55, 11), 'counter': 0L, 'active': True, 'provider_id': u'7js9mb', 'type': 1L, 'type_name': u'Visa', 'truncated_id': u'1111'}, 'transaction': {'amount': Decimal('10.00'), 'buyer': u'/generic/buyer/9/', 'carrier': None, 'created': datetime.datetime(2015, 6, 10, 17, 55, 12, 114201), 'currency': u'USD', 'notes': None, 'pay_url': None, 'provider': 4, 'region': None, 'related': None, 'relations': [], 'resource_pk': 17L, 'resource_uri': u'/generic/transaction/17/', 'seller': u'/generic/seller/54/', 'seller_product': u'/generic/product/48/', 'source': None, 'status': 4, 'status_reason': u'submitted_for_settlement', 'type': 0, 'uid_pay': None, 'uid_support': u'9zb22y', 'uuid': u'27864f4d-a813-4b89-9806-98d75fe06328'}, 'subscription': {'resource_pk': 14L, 'resource_uri': u'/braintree/mozilla/subscription/14/', 'paymethod': u'/braintree/mozilla/paymethod/14/', 'seller_product': u'/generic/product/48/', u'id': 14L, 'created': datetime.datetime(2015, 6, 10, 17, 55, 12), 'modified': datetime.datetime(2015, 6, 10, 17, 55, 12), 'counter': 0L, 'active': True, 'provider_id': u'kht8rm'}}, 'braintree': {'kind': u'subscription_charged_successfully', 'next_billing_date': datetime.date(2015, 7, 10), 'billing_period_start_date': datetime.date(2015, 6, 10), 'next_billing_period_amount': Decimal('10.00'), 'billing_period_end_date': datetime.date(2015, 7, 9)}}

andymckay:
This will work, but after looking at storing the notes in the transaction, I feel the braintree_transaction table is need. I'm tracking that in mozilla/payments#74 and I'll do that there. But this one should be good to merge tomorrow as I improve on this.

lib/brains/tests/test_webhook.py
```diff
@@ -88,37 +107,87 @@ def test_post_auth_fails(self):
...
 
     def process(self, subscription):
         process = Processor(notification(subject=subscription))
         process.update_transactions(process.webhook.subscription,
                                     self.braintree_sub)
 
+    def test_data(self):
+        process = Processor(notification(
+            subject=subscription(), kind=self.kind))
+        process.process()
```
kumar303: since this is two lines of code now, might be a good excuse to replace it with ``self.process()``

lib/brains/webhooks.py
kumar303: Maybe just add a comment here that we're only getting the most recent one. Otherwise the code looks awkward.

kumar303: I think somewhere here we should loop through all transactions and log some info about them just so we know what we're missing when we try to figure out how this will work in practice.

kumar303: oh nevermind, I see there is some logging in the processor functions

kumar303:
r+wc

lib/brains/webhooks.py
andymckay: This code is removed in the later pull request.

lib/brains/tests/test_webhook.py
andymckay: yeah i probably can...

andymckay: will pull this in a later commit so i don't have to rebase 2 more commits.

PR #474, opened 2015-06-11
add in braintree transaction

andymckay:
* fixes mozilla/payments#74 
* makes the Namespaced object simpler and a bit more flexible
* add in braintree transaction endpoint
* catch the submitted for settlement webhook as well

* requires #472, so maybe read that first

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -286,66 +286,83 @@ which then passes it on to this endpoint. For more information see the
         {
           "mozilla": {
             "paymethod": {
-              "resource_pk": 2,
-              "resource_uri": "/braintree/mozilla/paymethod/2/",
-              "braintree_buyer": "/braintree/mozilla/buyer/2/",
-              "id": 2,
-              "created": "2015-06-09T22:50:59.326",
-              "modified": "2015-06-09T22:50:59.326",
+              "resource_pk": 1,
+              "resource_uri": "/braintree/mozilla/paymethod/1/",
+              "braintree_buyer": "/braintree/mozilla/buyer/1/",
+              "id": 1,
+              "created": "2015-06-11T12:45:26.451",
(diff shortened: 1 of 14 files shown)


docs/topics/braintree.rst
```diff
@@ -281,7 +281,133 @@ which then passes it on to this endpoint. For more information see the
...
+.. _braintree-transaction-label:
+
+Transaction
+-----------
+
+The webhook returns transaction details to solitude. Solitude then creates a
+generic transaction object. It also creates a Braintree transaction that
+contains some information about the transaction.
+
+.. http:get:: /braintree/mozilla/subscription/<subscription id>/
```
kumar303: I'm confused. This was meant to be ``transaction``, right?

lib/brains/serializers.py
```diff
@@ -3,28 +3,30 @@
...
     @property
     def data(self):
-        return {
-            'mozilla': self.mozilla.data,
-            'braintree': self.braintree.data
-        }
+        def traverse(d):
+            for k, v in d.iteritems():
+                if isinstance(v, dict):
+                    traverse(v)
```
kumar303: ooh, fun times

lib/brains/serializers.py
```diff
@@ -66,6 +68,26 @@ def resource_uri(self, pk):
...
+
+    class Meta:
+        model = BraintreeTransaction
+        read_only_fields = (
+            'billing_period_end_date', 'billing_period_start_date', 'kind',
+            'next_billing_date', 'next_billing_period_amount'
+        )
+
+    def resource_uri(self, pk):
+        return reverse('generic:transaction-detail',
```
kumar303: this URL doesn't look correct

kumar303:
r+wc

lib/brains/serializers.py
andymckay: well spotted

PR #475, opened 2015-06-11
return the buyer

andymckay:
* add in the buyer object into the result so that the service can get the email

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -285,16 +285,30 @@ which then passes it on to this endpoint. For more information see the
 
         {
           "mozilla": {
+            "buyer": {
+              "active": true,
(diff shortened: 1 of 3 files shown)


docs/topics/braintree.rst
```diff
@@ -281,7 +281,148 @@ which then passes it on to this endpoint. For more information see the
...
     :<json bt_payload: the bt_payload issued by Braintree.
 
-    :status 204: webhook parsed successfully.
+    .. code-block:: json
+
+        {
+          "mozilla": {
+            "buyer": {
+              "active": true,
+              "email": null,
```
kumar303: for doc purposes you could put an email here. It won't be null.

kumar303:
r+wc

PR #476, opened 2015-06-11
use payments config

### LINKED ISSUE #454: Consume product information
Consume product information from mozilla/payments-config. So that we don't need to do as many look ups.

This is a speculative bug based on all the things we said we'd stick in here.

What does solitude need in payments-config?

andymckay:
* fixes #454 
* makes external_id and public_id the same, don't think that affects anything, that's a webpay thing


kumar303:
Looks right to me.

r+

PR #479, opened 2015-06-15
fake webhooks

andymckay:
fixes mozilla/payments#75

### DIFF
--- a/lib/brains/management/commands/braintree_webhook.py
+++ b/lib/brains/management/commands/braintree_webhook.py
@@ -1,0 +1,142 @@
+import base64
+import os
+from datetime import datetime, timedelta
+from optparse import make_option
+
+from django.core.management.base import BaseCommand, CommandError
+
+import braintree
+import requests
+from braintree.util.crypto import Crypto
+
+from lib.brains.models import BraintreeSubscription
+from lib.brains.tests.samples import webhooks
+from payments_config import products
+from solitude.logger import getLogger
+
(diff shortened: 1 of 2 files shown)


lib/brains/management/commands/braintree_webhook.py
```diff
@@ -0,0 +1,140 @@
...
+
+from lib.brains.models import BraintreeSubscription
+from lib.brains.tests.samples import webhooks
+from payments_config import products
+from solitude.logger import getLogger
+
+
+log = getLogger('s.brains.management')
+
+# These environment variables are populated by Docker. This script will not
```
kumar303: won't this code run even when this command doesn't run because of how Django imports it to discover the command? I think you should move this to the ``handle()`` function so that it only runs when the command runs.

kumar303:
Why does this have a test sample but no test? Is it missing a file?

lib/brains/management/commands/braintree_webhook.py
andymckay: They will, that's not great. Will move.

andymckay:
I didn't bother writing tests, I'm not sure how long this is going to stick around. I'll move the samples file unless you think it needs tests.

kumar303:
looks like this is missing the babel module in requirements

kumar303:
r+wc

andymckay:
I got babel in a seperate pull request I didn't push - doh!

PR #482, opened 2015-06-16
add in the product

### LINKED ISSUE #481: Expose product public_id in webhook response
We need the product's ``public_id`` in the [webhook response](https://solitude.readthedocs.org/en/latest/topics/braintree.html#post--braintree-webhook-) so that we can look up configuration with [payments_config](https://github.com/mozilla/payments-config).

andymckay:
* ..and I'm finding I keep doing print_as_json to generate docs and this saves me a few lines
* fixes #481

### DIFF
--- a/solitude/base.py
+++ b/solitude/base.py
@@ -20,6 +20,7 @@ from rest_framework import serializers, status
 from rest_framework.response import Response
 from rest_framework.views import APIView
 from rest_framework.viewsets import GenericViewSet
+from rest_framework.utils.encoders import JSONEncoder
 
 from solitude.logger import getLogger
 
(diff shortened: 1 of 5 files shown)


solitude/base.py
```diff
@@ -151,6 +152,13 @@ def mozilla_error(self, content, field):
     def braintree_error(self, content, field):
         return [f['code'] for f in content['braintree'][field]]
 
+    def print_as_json(self, data):
+        """
+        A utility function used to dump the json to stdout. Useful in tests
+        when you want some output for documentation.
+        """
+        print json.dumps(data, indent=2, cls=JSONEncoder)
```
kumar303: I found it really handy to copy/paste some json from the docs into a test but I had to reformat it to four space indents because of flake8. I don't know if it will cause docs to scroll off the page but maybe set it to four spaces for that reason?

kumar303:
r+wc

solitude/base.py
andymckay: Curious, why is flake8 complaining about JSON? You would have to convert it to a dict, or have it as a string?

PR #483, opened 2015-06-17
add in unsuccessfully charged webhook

### LINKED ISSUE #430: Cope with past due notifications
When a monthly transaction has failed, we will be notified. We've got to cope with that and the resulting attempts.

andymckay:
* fixes #430 
* in management command, allows the transaction to be specified in, otherwise creates a new one
* send unsuccessful in the management command

### DIFF
--- a/lib/brains/management/commands/braintree_webhook.py
+++ b/lib/brains/management/commands/braintree_webhook.py
@@ -1,5 +1,6 @@
 import base64
 import os
+import uuid
 from datetime import datetime, timedelta
 from optparse import make_option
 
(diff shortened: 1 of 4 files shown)


lib/brains/management/commands/braintree_webhook.py
```diff
@@ -92,8 +125,11 @@ def webhook(self, url, kind, sub):
             # Braintree doesn't assume months are 30 days long.
             'paid': datetime.today() + timedelta(days=29),
             'next': datetime.today() + timedelta(days=30),
+            'transaction': transaction[kind],
+            'processor-response': processor_response[kind]
```
andymckay: I changed this and corresponding keys to `processor_response` in a later pull request.

lib/brains/management/commands/braintree_webhook.py
```diff
@@ -68,6 +73,12 @@ class Command(BaseCommand):
             '--subscription_id',
             dest='subscription_id',
             help=('Primary key of the subscription in solitude.')
+        ),
+        make_option(
+            '--transaction_id',
```
kumar303: cool!

kumar303:
r+

Issue #462, opened 2015-06-05
Cope with a subscription_canceled web hook


andymckay-limited-access:
Looks like this:

```<WebhookNotification {timestamp: datetime.datetime(2015, 6, 5, 18, 24, 24), kind: u'subscription_canceled', subject: {u'subscription': {u'billing_period_end_date': None, u'updated_at': datetime.datetime(2015, 6, 5, 18, 24, 24), u'paid_through_date': None, u'id': u'fm2c3r', u'payment_method_token': u'733vmb', u'plan_id': u'mozilla-concrete-brick', u'next_billing_period_amount': u'10.00', u'failure_count': 0, u'merchant_account_id': u'xh6ry2b9dzjnjwtm', u'next_billing_date': datetime.date(2015, 7, 3), u'discounts': [], u'first_billing_date': datetime.date(2015, 7, 3), u'billing_period_start_date': None, u'number_of_billing_cycles': None, u'status': u'Canceled', u'billing_day_of_month': 3, u'transactions': [], u'current_billing_cycle': 0, u'price': u'10.00', u'trial_period': False, u'trial_duration_unit': None, u'add_ons': [], u'days_past_due': None, u'never_expires': True, u'created_at': datetime.datetime(2015, 6, 5, 10, 27, 56), u'status_history': [{u'status': u'Canceled', u'timestamp': datetime.datetime(2015, 6, 5, 18, 24, 24), u'price': u'10.00', u'user': u'amckay+braintree-sandbox-testing', u'subscription_source': u'control_panel', u'balance': u'0.00'}, {u'status': u'Pending', u'timestamp': datetime.datetime(2015, 6, 5, 10, 27, 56), u'price': u'10.00', u'user': u'amckay+braintree-sandbox-testing', u'subscription_source': u'api', u'balance': u'0.00'}], u'balance': u'0.00', u'trial_duration': None}}} at 140171227924752>```

**Comment by andymckay-limited-access** [comment]
Here's one with transactions:

```
<WebhookNotification {timestamp: datetime.datetime(2015, 6, 5, 18, 14, 39), kind: u'subscription_canceled', su
bject: {u'subscription': {u'billing_period_end_date': datetime.date(2015, 7, 2), u'updated_at': datetime.datetime(2015, 6, 5, 18, 14, 39), u'paid_through_da
[...]

PR #485, opened 2015-06-17
cope with subscription cancelled

### LINKED ISSUE #484: Cope with cancelled notifications
We've already got most of this, but make sure that the management command does this and something sane is returned to payments service.

andymckay:
* update the braintree_config command to make subscription_canceled possible
* fix the spelling of subscription_canceled in the code (bah!)
* fixes #484 
* cope with the fact that there may not be transactions in the response (see #462)
* save the subscription onto the Process object to stop it being passed around

### DIFF
--- a/lib/brains/management/commands/braintree_webhook.py
+++ b/lib/brains/management/commands/braintree_webhook.py
@@ -19,7 +19,8 @@ from solitude.logger import getLogger
 log = getLogger('s.brains.management')
 valid_kinds = [
     'subscription_charged_successfully',
-    'subscription_charged_unsuccessfully'
+    'subscription_charged_unsuccessfully',
+    'subscription_canceled',
 ]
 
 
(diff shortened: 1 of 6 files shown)


lib/brains/management/commands/braintree_webhook.py
```diff
@@ -19,7 +19,8 @@
 log = getLogger('s.brains.management')
 valid_kinds = [
     'subscription_charged_successfully',
-    'subscription_charged_unsuccessfully'
+    'subscription_charged_unsuccessfully',
+    'subscription_canceled'
```
kumar303: nit: trailing commas are nice for copy/paste

lib/brains/management/commands/samples/webhooks.py
```diff
@@ -193,3 +193,58 @@
   </subject>
 </notification>
 """  # noqa
+
+no_trans = """<?xml version="1.0" encoding="UTF-8"?>
```
kumar303: mmm, more XML. This will be not fun to maintain. I can't offer a better idea though.

lib/brains/tests/test_webhook.py
```diff
@@ -321,7 +334,7 @@ def test_ok(self):
 
 
 class TestWebhookSubscriptionCancelled(SubscriptionTest):
-    kind = 'subscription_cancelled'
+    kind = 'subscription_canceled'
```
kumar303: man, English is a messed up language

kumar303:
A quick update to the docs saying that the transaction might be null in the response would be cool.

r+wc

andymckay:
Yeah good call on updating docs.

lib/brains/management/commands/braintree_webhook.py
andymckay: JSON is ruining my life.

PR #486, opened 2015-06-18
Clear transactions for easier webhook simulation

kumar303:


andymckay:
Good idea, I think we should add a --clear-all or something that just does all the things, it's getting long to type.

r+

kumar303:
yeah, clearing subscriptions used to imply clear-all because of cascades.

PR #488, opened 2015-06-18
make descriptor better

andymckay:
* fixes #110 
* also some autoclean fixes
* also, what was that product.py file doing?


lib/brains/forms.py
```diff
@@ -97,16 +98,23 @@ def clean_plan(self):
         self.seller_product = obj
         return data
 
+    def get_name(self, plan_id):
+        # For detailed information on descriptors see: ttp://bit.ly/1CfDiou
```
kumar303: missing the h in http

lib/brains/tests/test_subscription.py
```diff
@@ -56,7 +56,7 @@ def test_ok(self):
             'payment_method_token': mock.ANY,
             'plan_id': 'brick',
             'descriptor': {
-                'name': 'Mozilla*product',
+                'name': 'Mozilla',
```
kumar303: did you try a subscription in the sandbox with this? I know it sounds weird but I swear I got an error when it didn't have a ``*`` symbol.

lib/brains/forms.py
```diff
@@ -97,16 +98,23 @@ def clean_plan(self):
...
         return {
             'payment_method_token': self.cleaned_data['paymethod'].provider_id,
-            'plan_id': self.seller_product.public_id,
+            'plan_id': plan_id,
             'trial_period': False,
             'descriptor': {
-                # TODO: figure out how to get product in here
-                # https://github.com/mozilla/payments/issues/57
-                'name': 'Mozilla*product',
+                'name': self.get_name(plan_id),
```
kumar303: This has to follow a [strict format](https://developers.braintreepayments.com/javascript+python/reference/response/subscription#descriptor.name) so we should do some truncating since we're sourcing it from configured product names. It's a little hard to interpret but I think it needs to be ``Mozilla*something`` and the total needs to be no longer than 22 characters. That means we probably want to truncate the product like ``Mozilla*Concretexxxxxx`` (imagine the X's are actual letters).

kumar303:
r-, pretty sure this needs to adhere to the documented format. However, the docs are a little confusing so I could be wrong.

andymckay:
Poopy. But the error you get back is actually descriptive:

Error on creating subscription: Company name/DBA section must be either 3, 7 or 12 characters and the product descriptor can be up to 18, 14, or 9 characters respectively (with an * in between for a total descriptor name of 22 characters)

kumar303:
yeah, same thing I linked to

lib/brains/forms.py
```diff
@@ -97,16 +98,30 @@ def clean_plan(self):
...
         return data
 
+    def format_descriptor(self, name):
+        # The rules for descriptor are:
+        #
+        # Company name/DBA section must be either 3, 7 or 12 characters and
+        # the product descriptor can be up to 18, 14, or 9 characters
+        # respectively (with an * in between for a total descriptor
+        # iname of 22 characters)
+        return 'Mozilla*{}'.format(name.replace(' ', ''))[0:22]
```
kumar303: it might be nice to log this so we can check back to see if something is nonsensical

lib/brains/forms.py
```diff
@@ -97,16 +98,30 @@ def clean_plan(self):
...
+
+    def get_name(self, plan_id):
+        if plan_id in products:
+            return unicode(products.get(plan_id).description)
```
kumar303: why coerce to unicode here?

kumar303: oh, sigh, because of the magic localizer, nevermind

lib/brains/forms.py
```diff
@@ -97,16 +98,30 @@ def clean_plan(self):
...
+        log.warning('Unknown product for descriptor: {}'.format(plan_id))
+        return 'Product'
```
kumar303: maybe default to ``Service`` ? Still vague but possibly more useful.

lib/brains/tests/test_forms.py
```diff
@@ -45,3 +46,23 @@ def test_parse_ok(self):
...
+    def test_seller_name(self):
+        eq_(self.form.get_name('mozilla-concrete-brick'), 'Brick')
+
+    def test_format_descriptor(self):
+        for in_string, out_string in [
+                ('Product', 'Mozilla*Product'),
+                ('With Space', 'Mozilla*WithSpace'),
+                ('Long Product With Space', 'Mozilla*LongProductWit')
+            ]:
+            eq_(self.form.format_descriptor(in_string), out_string)
```
kumar303: nice tests

lib/brains/forms.py
andymckay: /me shrugs, should never happen.

kumar303:
r+wc

PR #489, opened 2015-06-20
upgrade m2crypto and remove sudo

andymckay:


andymckay:
And docker builds just fine as well.

andymckay:
Note to future self: this works because pyrepo has a wheel of M2Crypto which means that travis doesn't need to build it.

PR #493, opened 2015-06-30
delete payment methods from braintree

### LINKED ISSUE #492: Delete pay method
When a pay method is made inactive in solitude,  or soft deleted, do a delete in Braintree.

https://developers.braintreepayments.com/javascript+python/reference/request/payment-method/delete

### LINKED ISSUE #494: Prevent re-activating payment methods
Once a payment method is set to active=False, it cannot be set to active=True, because we've deleted it from Braintree and there isn't an undelete option.

andymckay:
* calls delete upon changing status to active
* if this happens multiple times it doesn't fail
* fixes #492 and supports mozilla/payments#84
* also prevents enabling a deactivated payment method which fixes #494

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -3,7 +3,10 @@ from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 
 import requests
+from braintree.exceptions.not_found_error import NotFoundError
 
+from lib.brains.client import get_client
+from lib.brains.errors import BraintreeResultError
 from lib.brains.models import BraintreeBuyer, BraintreePaymentMethod
 from lib.buyers.models import Buyer
 from lib.sellers.models import SellerProduct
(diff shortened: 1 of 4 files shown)


lib/brains/forms.py
```diff
@@ -167,3 +169,44 @@ def clean(self):
...
+        self.object = obj
+        return super(PayMethodUpdateForm, self).__init__(data)
+
+    def clean_active(self):
+        active = self.cleaned_data['active']
+
+        # An attempt to enable active on the payment method.
+        if active and not self.object.active:
+            raise forms.ValidationError(
+                'Cannot set a payment method to active', code='invalid')
```
muffinresearch: "Cannot set an inactive payment method to active" or similar might be tiny bit clearer.

muffinresearch:
r+wc

PR #500, opened 2015-07-06
Cancel subscription

andymckay:
* won't allow a subscription to be reactivated
* cancels the subscription in braintree

```
curl --dump-header - -H "Content-Type: application/json" -X POST --data '{"subscription": "/braintree/mozilla/subscription/1/"} http://solitude:2602/braintree/subscription/cancel/
```

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -232,11 +232,19 @@ Change payment method on a subscription:
 .. http:post:: /braintree/subscription/paymethod/change/
 
     :<json string paymethod: the resource_uri of the paymethod in solitude.
+
+    :status 200: subscription changed.
+
+Cancel a subscription. This will cancel the subscription in Braintree and is
+not reversible.
+
+.. http:post:: /braintree/subscription/cancel/
+
     :<json string subscription: the resource_uri of the subscription in solitude.
 
     The response is in the same format as for creation.
 
-    :status 200: subscription changed.
+    :status 200: subscription cancelled.
 
 Data stored in solitude
 +++++++++++++++++++++++
--- a/lib/brains/tests/test_subscription.py
+++ b/lib/brains/tests/test_subscription.py
@@ -11,7 +11,6 @@ from nose.tools import eq_, ok_
 from lib.brains.models import BraintreeSubscription
 from lib.brains.tests.base import (
     BraintreeTest, create_braintree_buyer, create_method, create_seller, error)
-from solitude.base import APITest
 
 
 def method(**kw):
@@ -24,7 +23,7 @@ def method(**kw):
     return Subscription(None, method)
 
 
-def successful_method(**kw):
+def successful_subscription(**kw):
     return SuccessfulResult({'subscription': method(**kw)})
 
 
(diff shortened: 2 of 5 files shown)


docs/topics/braintree.rst
```diff
@@ -227,6 +227,17 @@ solitude.
 
     :status 201: payment method created.
 
+Cancel a subscription. This will cancel the subscription in Braintree and is
+not reversible.
+
+.. http:post:: /braintree/subscription/cancel/
+
+    :<json string paymethod: the resource_uri of the subscription in solitude.
```
kumar303: I hope the parameter to this really isn't ``paymethod`` :fries:

kumar303: I'm still thinking that a PUT on the subscription resource would make sense here. There are multiple data objects but that's just an implementation detail. To the caller, there is only one subscription by one ID.

kumar303: Another possibility would be to do ``POST /braintree/subscription/123/cancel/`` which wouldn't require passing in a ``resource_uri``

lib/brains/tests/test_subscription.py
```diff
@@ -35,18 +34,18 @@ def create_method_all():
...
-        super(TestSubscriptionMethod, self).setUp()
+        super(TestSubscription, self).setUp()
         self.url = reverse('braintree:subscription')
 
     def test_allowed(self):
         self.allowed_verbs(self.url, ['post'])
 
     def test_ok(self):
-        self.mocks['sub'].create.return_value = successful_method()
+        self.mocks['sub'].create.return_value = successful_subscription()
```
kumar303: thanks for fixing up these names

lib/brains/tests/test_subscription.py
```diff
@@ -98,7 +97,56 @@ def test_braintree_fails(self):
...
+        self.mocks['sub'].cancel.return_value = error()
+
+        obj = self.create_subscription()
+        res = self.post(obj)
+        eq_(res.status_code, 422, res.content)
+        eq_(obj.reget().active, True)
+
+        self.mocks['sub'].cancel.assert_called_with(obj.provider_id)
+
+    def test_active_post(self):
```
kumar303: This would be better named as ``test_cannot_cancel_inactive()`` otherwise it needs a comment or something

kumar303:
I think this can be deleted from the docs too http://solitude.readthedocs.org/en/latest/topics/braintree.html#patch--braintree-mozilla-subscription--subscription%20id--

kumar303:
r+wc

docs/topics/braintree.rst
andymckay-limited-access: doc copy paste fail, thanks

andymckay-limited-access: a URL like that does require you making up a special URL though, i'm going to make the assumption that the URL to all the endpoints are stored in payments-service and that code that just says pass these two existing variables is easier to actually implement.

andymckay-limited-access:
will update docs, thanks

andymckay:
well the mozilla data in solitude is still patchable... will need to think if active should be patchable, i'm starting to think that solitude needs to do more with authorization - here are the thing payments-service is authorized to access

PR #501, opened 2015-07-06
improved pay method

andymckay:
* proposed change to paymethod cancelling, just an explicit end point
* seperates braintree and mozillay
Now to cancel:

```
curl --dump-header - -H "Content-Type: application/json" -X POST --data '{"paymethod": "/braintree/mozilla/paymethod/1/"} http://solitude:2602/braintree/paymethod/cancel/
```

* this would mean removing the ability to patch /braintree/paymethod/1/ from the bodyguard and changing to this.

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -99,12 +99,12 @@ Some information is stored in solitude after creating a customer.
 Payment Methods
 ---------------
 
-Creates a payment method in Braintree and the corresponding payment method in
-solitude.
+Create or update a payment method in Braintree and the corresponding payment
+method in solitude.
 
 .. http:post:: /braintree/paymethod/
 
-    :<json string uuid: the uuid of the buyer in solitude.
+    :<json string buyer_uuid: the uuid of the buyer in solitude.
     :<json string nonce: the payment nonce returned by Braintree.
 
     .. code-block:: json
@@ -131,11 +131,22 @@ solitude.
             }
         }
 
-    :>json string braintree token: id of the payment method in braintree.
+    :>json string braintree token: id of the payment method in Braintree.
     :>json string braintree created_at: created date and time.
     :>json string braintree updated_at: updated date and time.
 
     :status 201: payment method created.
+
+Delete a payment method. This will delete the payment method in Braintree
+and is not reversible.
+
+.. http:post:: /braintree/paymethod/delete/
+
+    :<json string paymethod: the resource_uri of the payment method in solitude.
+
+    The response is in the same format as for creation.
+
+    :status 200: payment method cancelled.
 
 Data stored in solitude
 +++++++++++++++++++++++
@@ -168,10 +179,6 @@ Some information about the payment method is stored in solitude.
                                 for a credit card, the last 4 digits, this field is read only.
     :>json int type: `1` for credit card is currently the only one supported, this field is read only.
     :>json string type_name: name of the type of purchase, this field is read only.
-
-.. http:patch:: /braintree/mozilla/paymethod/<method id>/
-
-    :<json boolean active: if the payment method is currently active.
 
 .. http:get:: /braintree/mozilla/paymethod/
 
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -3,10 +3,7 @@ from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 
 import requests
-from braintree.exceptions.not_found_error import NotFoundError
 
-from lib.brains.client import get_client
-from lib.brains.errors import BraintreeResultError
 from lib.brains.models import BraintreeBuyer, BraintreePaymentMethod
 from lib.buyers.models import Buyer
 from lib.sellers.models import SellerProduct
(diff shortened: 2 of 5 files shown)


lib/brains/views/paymethod.py
```diff
@@ -14,55 +14,73 @@
...
-    solitude_method = BraintreePaymentMethod.objects.create(
-        braintree_buyer=braintree_buyer,
-        type=PAYMENT_METHOD_CARD,
-        type_name=braintree_method.card_type,
-        provider_id=braintree_method.token,
-        truncated_id=result.payment_method.last_4
-    )
-    log.info('Method {0} created.'.format(solitude_method.pk))
+        client = get_client().PaymentMethod
+        result = client.delete(solitude_method.provider_id)
```
kumar303: shouldn't this only delete from Braintree when the patch is setting it to inactive? Or maybe this should be in a ``delete`` handler instead of patch?

kumar303:
I think I need to come back to this patch after another revision. Here are some thoughts:

* I think none of the buyer verification belongs in here. The ``service`` will have to verify that the buyer owns the payment method before allowing it to be deactivated/deleted. Solitude can't really do anything to protect against this.
* Thus, we should be able to simply patch the full URL of the resource instead of requiring that the resource URL be passed in as patched data
* I like the idea of separating out the braintree deletion into a separate call

lib/brains/views/paymethod.py
kumar303: ok, I guess the form only allows patches of ``active=False``?

andymckay: One of those, I think I forgot to check that. We could do a `delete` but deletes that don't really delete seem really bad too :)

lib/brains/forms.py
```diff
@@ -173,42 +170,31 @@ def clean(self):
...
+            raise forms.ValidationError(
+                'Paymethod is required', code='required'
+            )
+
+        if (self.cleaned_data.get('buyer_uuid') !=
+                solitude_method.braintree_buyer.buyer.uuid):
+            # Can I raise a 403 from here?
+            raise forms.ValidationError(
+                'Wrong user', code='invalid'
+            )
```
andymckay: As per chat in IRC @kumar303 do you think this is of any value?

andymckay:
Sorry, wasn't quite ready for a full r? but at the "what do we think of this" stage. Will give it a clean up tomorrow.

andymckay:
Sorry, wasn't quite ready for a full r? but at the "what do we think of this" stage. Will give it a clean up tomorrow.

I'm not sure what you mean "simply patch the full URL of the resource", because that conflicts with "I like the idea of separating out the braintree deletion into a separate call"... if I understand your meaning of the full URL of the resource correctly.

kumar303:
Let me know when it's ready for another look and perhaps I can give some more cohesive feedback

docs/topics/braintree.rst
```diff
@@ -131,12 +131,23 @@ solitude.
...
+    :>json string braintree token: id of the payment method in Braintree.
     :>json string braintree created_at: created date and time.
     :>json string braintree updated_at: updated date and time.
 
     :status 201: payment method created.
 
+To cancel a payment method. This will delete the payment method in Braintree
+and is not reversible.
+
+.. http:post:: /braintree/paymethod/cancel/
```
kumar303: Why cancel? That's confusing to me because you'd cancel a subscription which is different but related. The user would want to delete a payment method. Maybe ``/braintree/paymethod/delete/`` or ``/braintree/paymethod/remove/`` ?

andymckay: Sure

docs/topics/braintree.rst
```diff
@@ -131,12 +131,23 @@ solitude.
...
+
+    :<json string paymethod: the resource_uri of the payment method in solitude.
+
+    The response is in the same format as for creation.
```
kumar303: This seems weird to me because the response has created_at timestamps https://solitude.readthedocs.org/en/latest/topics/braintree.html#post--braintree-paymethod-

lib/brains/forms.py
```diff
@@ -172,43 +169,22 @@ def clean(self):
...
+        if not solitude_method:
+            raise forms.ValidationError(
+                'Paymethod is required', code='required'
+            )
 
         # An attempt to enable active on the payment method.
-        if active and not self.object.active:
+        if not solitude_method.active:
             raise forms.ValidationError(
                 'Cannot set an inactive payment method to active',
```
kumar303: This doesn't make sense anymore. If this form is only for removing payment methods, how would it ever activate a payment method?

lib/brains/views/paymethod.py
```diff
@@ -15,6 +15,36 @@
...
+
+    solitude_method.active = False
+    solitude_method.save()
+
+    log.info('Payment method deleted from braintree: {}'
+             .format(solitude_method.pk))
+
+    res = serializers.Namespaced(
+        mozilla=serializers.LocalPayMethod(instance=solitude_method),
+        braintree=serializers.PayMethod(instance=result.payment_method)
```
kumar303: what is ``result.payment_method`` when you delete it?

docs/topics/braintree.rst
kumar303: or maybe ``/braintree/paymethod/deactivate/`` ?

kumar303: I still think ``DELETE /braintree/paymethod/123/`` would make sense here.

andymckay: Just grumpy that its not a delete.

lib/brains/views/paymethod.py
andymckay: Its information about the payment_method, same as on a create.

docs/topics/braintree.rst
andymckay: The pay method was still created.

docs/topics/braintree.rst
kumar303: But is a delete ever a delete? ;)

kumar303:
r+wc

PR #502, opened 2015-07-08
change a subscriptions paymethod

andymckay:
* won't allow you to use an inactive subscription or paymethod

```
curl --dump-header - -H "Content-Type: application/json" -X POST --data '{"subscription": "/braintree/mozilla/subscription/2/", "paymethod": "/braintree/mozilla/paymethod/1/"}' http://solitude:2602/braintree/subscription/change/
```

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -227,6 +227,16 @@ solitude.
 
     :status 201: payment method created.
 
+Change payment method on a subscription:
+
+.. http:post:: /braintree/subscription/paymethod/change/
+
+    :<json string paymethod: the resource_uri of the paymethod in solitude.
+    :<json string subscription: the resource_uri of the subscription in solitude.
+
+    The response is in the same format as for creation.
+
+    :status 200: subscription changed.
 
 Data stored in solitude
 +++++++++++++++++++++++
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -4,7 +4,8 @@ from django.core.exceptions import ObjectDoesNotExist
 
 import requests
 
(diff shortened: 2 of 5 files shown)


docs/topics/braintree.rst
```diff
@@ -227,6 +227,16 @@ solitude.
 
     :status 201: payment method created.
 
+Change payment method on a subscription:
+
+.. http:post:: /braintree/subscription/change/
```
kumar303: ``change`` is pretty vague if this is only allowing the payment method to be updated. I guess we might add more later? Otherwise ``/braintree/subscription/paymethod/`` seems fine to me.

kumar303:
I'm still not totally clear why a ``PUT`` on ``subscription['resource_uri']`` is unacceptable here. It would make the API a lot more intuitive. Even though the data is stored in two places, it's still a single object as far as the API caller is concerned.

kumar303:
r+wc

docs/topics/braintree.rst
kumar303: If you are still opposed to PUT, another thing that might make this more intuitive is to put the ID in the resource, like: ``/braintree/subscription/123/change/`` because otherwise it seems odd to me that you'd have to POST the ID to the subscription.

andymckay: what data do you have in the payments-service, the URL to the solitude subscription?

andymckay:
intuitive APIs that do magic have been the bane of solitudes existence as we try to hide the complexity

docs/topics/braintree.rst
andymckay: true, change is vague

Issue #503, opened 2015-07-08
Shorter transaction references


andymckay-limited-access:
The transaction references sent on receipts are uuids, which is 36 characters. Rather unwieldy, we can probably do better than that and get something a little shorter, maybe just re-use the braintree ones.

PR #505, opened 2015-07-08
shorter transaction id based on the pk

andymckay:
* currently 40000 so it will look like this, up to...

```
In [4]: shorter(40000)
Out[4]: 'nUs'

In [5]: shorter(4000000000000)
Out[5]: 'ctRdH6eJ'
```

### DIFF
--- a/lib/brains/webhooks.py
+++ b/lib/brains/webhooks.py
@@ -1,4 +1,4 @@
-import uuid
+import time
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
@@ -8,6 +8,7 @@ from lib.brains.serializers import serialize_webhook
 from lib.transactions import constants
 from lib.transactions.models import Transaction
 from solitude.base import getLogger
+from solitude.utils import shorter
 
 log = getLogger('s.webhooks')
 
@@ -228,9 +229,21 @@ class Processor(object):
                     status=our_status,
                     status_reason=reason,
                     type=constants.TYPE_PAYMENT,
-                    uid_support=their_transaction.id,
-                    uuid=str(uuid.uuid4())
+                    uid_support=their_transaction.id
                 )
+
+                # Make the transaction short and helpful
+                # (14 chars, still 4 chars less than the mocks)
+                our_transaction.uuid = (
+                    'bt-{}-{}'.format(
+                        # Should be unique within this db.
+                        shorter(our_transaction.pk),
+                        # If pk is repeated (eg. in tests, dev) should still
+                        # be unique.
+                        shorter(int(time.time())))
+                    )
+                our_transaction.save()
+
                 log.info('Transaction created: {}'.format(our_transaction.pk))
 
                 braintree_transaction = BraintreeTransaction.objects.create(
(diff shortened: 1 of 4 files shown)


andymckay:
If someone reads the code, this does expose the pk to people. I'm not sure if thats a problem though really.

lib/brains/webhooks.py
```diff
@@ -228,9 +227,13 @@ def update_transactions(self):
...
                     status_reason=reason,
                     type=constants.TYPE_PAYMENT,
-                    uid_support=their_transaction.id,
-                    uuid=str(uuid.uuid4())
+                    uid_support=their_transaction.id
                 )
+
+                # Make the transaction short and helpful.
+                our_transaction.uuid = 'bt-' + shorter(our_transaction.pk)
+                our_transaction.save()
```
kumar303: it's a little unfortunate that we have to hit the db twice here but it's only a webhook so probably not a big deal.

solitude/utils.py
```diff
@@ -2,6 +2,22 @@
 from django.core.exceptions import ImproperlyConfigured
 
 
+# Note deliberately ignoring o, O, 0, i and 1 as too similar.
```
kumar303: might want to ignore `l` (lowercase L) while you're at it since that looks like a 1

solitude/utils.py
```diff
@@ -2,6 +2,22 @@
...
+
+
+def shorter(integer):
+    """
+    Convert an integer into something shorter.
+    """
+    output = ''
+    while integer > 0:
+        remainder = integer % len(choices)
+        output = choices[remainder] + output
```
kumar303: if you did something more like:
````
output = []
output.append(choices[remainder])
return ''.join(output)
````
then I think it would be noticeably faster for bigger numbers. Same amount of code.

File: solitude/tests/test.py [hunk 4]
```diff
@@ -94,3 +95,9 @@ def test_ok(self):
     def test_not_ok(self):
         self.req.QUERY_PARAMS = {'uid': ['bar']}  # Note the typo there.
         StrictQueryFilter().filter_queryset(self.req, self.queryset, self.view)
+
+
+class TestShorter(APITest):
```
kumar303: could just be ``unittest.TestCase``

solitude/utils.py
```diff
@@ -2,6 +2,22 @@
...
+        integer /= len(choices)
+    return output
```
kumar303: that's pretty cool

kumar303:
r+wc

lib/brains/webhooks.py
andymckay: Yeah I couldn't think of a way to do it based on the pk without the extra hit and using the pk seems like an easy way to ensure uniqueness.

solitude/utils.py
andymckay: good idea

solitude/utils.py
andymckay: Sure although how much of an improvement it is anymore it certainly has changed over the Python versions. https://gist.github.com/andymckay/8a93550b725b1abba270. http://stackoverflow.com/questions/1349311/python-string-join-is-faster-than-but-whats-wrong-here is a good read.

andymckay-limited-access:
fixes #503

Issue #497, opened 2015-07-06
Support delete account


andymckay:
To be completely determined what happens on delete account, but when someone hits delete account in the payments-ui, we should:
* cancel all subscriptions
* cancel all payment-methods
* randomise the users FxA id and email address

Leaving all the data there for stats and referential integrity but no longer tied to the user.

Supports mozilla/payments#85

andymckay-limited-access:
The model code is there, will just need hooking up with a REST endpoint in solitude.

PR #506, opened 2015-07-10
provide a close buyer method on the model

andymckay:
* will close the issue when the API is accessible and documented from a HTTP request
* just the internal stuff for the moment

### DIFF
--- a/lib/brains/models.py
+++ b/lib/brains/models.py
@@ -1,9 +1,14 @@
 from django.core.urlresolvers import reverse
 from django.db import models
+from django.dispatch import receiver
 
+from lib.brains.client import get_client
+from lib.brains.errors import BraintreeResultError
 from lib.buyers.models import Buyer
-from solitude.base import Model
+from solitude.base import getLogger, Model
 from solitude.constants import PAYMENT_METHOD_CARD
+
+log = getLogger('s.brains')
 
 
 class BraintreeBuyer(Model):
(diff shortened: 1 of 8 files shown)


andymckay-limited-access:
fixes #497

lib/brains/models.py
```diff
@@ -24,12 +29,41 @@ def get_uri(self):
...
+        # case continue.
+        log.info('No braintree buyer found for buyer: {}'
+                 .format(buyer.pk))
+        return
+
+    for paymethod in braintree_buyer.paymethods.all():
+
+        # Find and clear out all subscriptions.
+        for subscription in paymethod.subscriptions.filter(active=True):
+            subscription.braintree_cancel()
```
kumar303: I guess this will send an email, right? If someone is deleting their account will they want a bunch of emails for each of their cancelled subscriptions?

lib/brains/tests/test_models.py
```diff
@@ -0,0 +1,65 @@
...
+
+        eq_(self.method.reget().active, False)
+        eq_(self.sub.reget().active, False)
+
+    def test_listens_signal(self):
+        self.mocks['pay'].delete.return_value = successful_method()
+        self.mocks['sub'].cancel.return_value = successful_subscription()
+
+        self.buyer.close_signal.send(
+            buyer=self.buyer, sender=self.buyer.__class__)
```
kumar303: do the mocks need call assertions or does that happen automatically? I forget.

lib/brains/tests/test_models.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_inactive_method(self):
+        # If a method is inactive, then we still go and test its payment
+        # methods.
```
kumar303: this comment doesn't make sense to me. Test its payment methods? Did you mean subscription?

lib/brains/models.py
andymckay: Currently it call the braintree API, which will then trigger the webhooks, which will then come back to payment-service, hit solitude and generate an email. So yes, they'll get an email. Unless you've got a ton of accounts, I still kind of like that idea of getting the confirmation by email.

kumar303: yeah, I guess sending email per subscription is a good idea

kumar303:
r+wc

lib/brains/tests/test_models.py
andymckay: it asserts an mock defined is called, so unless i'm specifically testing the params (done in the test above) there's no need to test.

lib/brains/tests/test_models.py
andymckay: ugh wasn't the best comment, let me reword that

PR #507, opened 2015-07-13
run braintree_config automatically if env exists

andymckay:


bin/docker_run.sh
```diff
@@ -12,4 +12,9 @@ if [ $? -ne 0 ]; then
     schematic migrations/
 fi
 
+if [ "$BRAINTREE_MERCHANT_ID" ]; then
+    echo "Braintree merchant ID found in environment, running braintree_config."
+    python manage.py braintree_config
```
kumar303: are you sure this can be run repetitively? I got this when I ran it: `django.db.utils.IntegrityError: (1062, "Duplicate entry 'mozilla-concrete-brick' for key 'public_id'")`

bin/docker_run.sh
```diff
@@ -12,4 +12,9 @@ if [ $? -ne 0 ]; then
     schematic migrations/
 fi
 
+if [ "$BRAINTREE_MERCHANT_ID" ]; then
```
kumar303: bash is pretty cryptic but shouldn't this be ``if [ -z "$BRAINTREE_MERCHANT_ID" ]`` ?

bin/docker_run.sh
andymckay: It sure can. Be curious why that would be the case on your solitude db.

bin/docker_run.sh
andymckay: I think, -z is "True of the length if "STRING" is zero." - http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

```
pay master payments-service $ cat foo.bash 
if [ "$FOO" ]; then
    echo "Found env."
fi
pay master payments-service $ env | grep FOO
pay master payments-service $ ./foo.bash 
pay master payments-service $ export FOO=
pay master payments-service $ ./foo.bash 
pay master payments-service $ export FOO=something
pay master payments-service $ ./foo.bash 
Found env.
```

bin/docker_run.sh
kumar303: hmm, it was probably an artifact of before we switched around the public_id. I deleted everything and was able to run it multiple times after that.

kumar303:
r+

PR #508, opened 2015-07-13
move to latest version of marketplace docker build

andymckay:
....also note that requirements/docs.txt has multiple conflicts with prod.txt and won't work on newer versions of pip. So removing that until someone fixes it up.


kumar303:
I'm sure ops will love `latest` ! jk. Much work still needed for prod-ready docker.

r+

PR #512, opened 2015-07-15
add in locale fixes #511

### LINKED ISSUE #511: Store users locale
When we create a buyer, store the users locale at that time. We'll use that to localize emails later on.

andymckay:


lib/buyers/tests/test_api.py
```diff
@@ -75,6 +76,7 @@ def test_get(self):
         eq_(data['pin_is_locked_out'], False)
         eq_(data['pin_was_locked_out'], False)
         eq_(data['email'], self.email)
+        eq_(data['locale'], 'en-US')
```
kumar303: maybe we should store everything in lower case?

docs/topics/generic.rst
```diff
@@ -82,6 +85,8 @@ You can also get the details of a buyer:
 
     :param email: users email address.
     :param type: string
+    :param locale: users locale, most likely the Accept Language HTTP header
```
kumar303: nit: user's locale...

migrations/58-add-locale.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `buyer` ADD COLUMN `locale` VARCHAR(255);
```
kumar303: this will be `NULL`, right ?

lib/buyers/tests/test_api.py
andymckay: let's leave that up to payments-service, it would be annoying if my data store lower cased things for me :)

migrations/58-add-locale.sql
andymckay: `null` is allowed unless you specify not `null`

kumar303:
r+

PR #513, opened 2015-07-15
pass through the correct header

andymckay:
for mozilla/payments#89

### DIFF
--- a/lib/provider/views.py
+++ b/lib/provider/views.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.core.urlresolvers import reverse
 
 from django_statsd.clients import statsd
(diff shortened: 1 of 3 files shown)


kumar303:
ouch, this is ugly. Are you sure this won't break bango? I think we ported some stuff to the abstract provider but maybe only in webpay.

r+ ... I guess? :fire: 

as long as nothing else breaks.

lib/provider/views.py
```diff
@@ -72,6 +73,16 @@ def error(self, response):
 
     def result(self, proxied_endpoint, *args, **kwargs):
         method = getattr(proxied_endpoint, '__name__', 'unknown_method')
+
+        # Not all test mocks have this. Someone more dedicated than me
+        # can fix up all the tests for this soon to be deleted feature.
+        if (hasattr(proxied_endpoint, '__self__')
```
kumar303: I don't really see why this is helpful. Why don't you just skip the tests instead? Are they even covering anything with this workaround code?

PR #514, opened 2015-07-17
pass the right URL

andymckay:
* i was passing through to the auth that the final destination of zippy... was solitude-auth, that's silly the final URL for zippy requests is zippy, that's what the header is for
* let's move the final url calculation into the proxy, that's what the current lib/proxy does and makes this code easier
* let's set the default url to be the docker one, not paas


kumar303:
verynice-1

not sure what the test failures are about but this looks wayyyyy better.

r+

PR #517, opened 2015-07-20
no need for solitude proxy any more

### LINKED ISSUE #515: Remove solitude-proxy from docker
The solitude proxy process within docker is no longer needed since the docker environments rely on solitude-auth for now.

andymckay:
* things go through solitude-auth now, so remove the second process inside the solitude container
* set the default in settings to solitude-auth for docker to save another env variable
* fixes #515

### DIFF
--- a/lib/provider/tests/test_views.py
+++ b/lib/provider/tests/test_views.py
@@ -28,9 +28,11 @@ class TestAPIView(TestCase):
 
     def test_no_reference(self):
         req = RequestFactory().get('/')
-        with self.settings(ZIPPY_MOCK=False):
-            eq_(ProxyView().dispatch(req, reference_name='bob',
-                                     resource_name='sellers').status_code, 404)
+        with self.assertRaises(KeyError):
+            with self.settings(ZIPPY_MOCK=False):
+                eq_(ProxyView().dispatch(
+                    req, reference_name='bob',
+                    resource_name='sellers').status_code, 404)
 
 
 class TestAPIasProxy(TestCase):
--- a/solitude/settings/base.py
+++ b/solitude/settings/base.py
@@ -477,7 +477,8 @@ ZIPPY_CONFIGURATION = {
 }
 
 # The URL for a solitude proxy to zippy.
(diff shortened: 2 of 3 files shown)


kumar303:
looks like the tests might be missing some config

lib/provider/tests/test_views.py
```diff
@@ -28,9 +28,11 @@ class TestAPIView(TestCase):
 
     def test_no_reference(self):
         req = RequestFactory().get('/')
-        with self.settings(ZIPPY_MOCK=False):
-            eq_(ProxyView().dispatch(req, reference_name='bob',
-                                     resource_name='sellers').status_code, 404)
+        with self.assertRaises(KeyError):
```
kumar303: what is this test doing? Do we still need it? I can't decipher from the test name what it's testing.

lib/provider/tests/test_views.py
```diff
@@ -28,9 +28,11 @@ class TestAPIView(TestCase):
...
+            with self.settings(ZIPPY_MOCK=False):
+                eq_(ProxyView().dispatch(
+                    req, reference_name='bob',
+                    resource_name='sellers').status_code, 404)
```
kumar303: if you're just going for a KeyError maybe you don't need `eq_` ?

solitude/settings/base.py
```diff
@@ -477,7 +477,8 @@
 }
 
 # The URL for a solitude proxy to zippy.
-ZIPPY_PROXY = os.getenv('SOLITUDE_ZIPPY_PROXY', '')
+ZIPPY_PROXY = os.getenv(
+    'SOLITUDE_ZIPPY_PROXY', 'http://solitude-auth:2603/v1/reference/')
```
kumar303: I think it's a little nicer to account for empty variable values which sometimes happens in docker setups. This would do it:
````
os.getenv('SOLITUDE_ZIPPY_PROXY') or 'http://solitude-auth:2603/v1/reference/'
````

kumar303:
r+wc

PR #519, opened 2015-08-03
Fix return result of pay method deletion

kumar303:
The braintree result does not contain the payment_method definition.


muffinresearch:
lgtm r+

kumar303:
needed for https://github.com/mozilla/payments-service/issues/89

PR #521, opened 2015-08-06
Set SITE URL as a variable for the payments-proxy

lazaroj:
From Bug 1191932:

Greetings,

We'd like to set the SITE_URL as a variable for payments-proxy.  I've created a PR here, which seems like a small change, but wanted to verify with the dev team that this would not impact anything else.

The current SITE_URL for prod is set to https://payments-proxy.firefox.com, where we would want to configure that to https://payments-proxy.prod.mozaws.net as a variable instead of a static entry.

Thanks!


andymckay:
r+

PR #522, opened 2015-08-06
turn on atomic requests on prod

### LINKED ISSUE #520: Rollback error occuring on /generic/buyer/
It looks like webpay is creating users and that is triggering an error.

```
Aug  5 07:11:44 localhost:  s.auth:INFO webpay:- Access granted for: webpay, to: /generic/buyer/ :./solitude/authen
tication.py:56
Aug  5 07:11:44 localhost:  django.request:ERROR : Internal Server Error: /generic/buyer/ :/data/sol.prod/www/payme
nts.firefox.com/current/venv/lib/python2.7/site-packages/django/core/handlers/base.py:256#012Traceback (most recent
 call last):#012  File "/data/sol.prod/www/payments.firefox.com/current/venv/lib/python2.7/site-packages/django/cor
e/handlers/base.py", line 132, in get_response#012    response = wrapped_callback(request, *callback_args, **callba
ck_kwargs)#012  File "/data/sol.prod/www/payments.firefox.com/current/venv/lib/python2.7/site-packages/rest_framewo
rk/viewsets.py", line 78, in view#012    return self.dispatch(request, *args, **kwargs)#012  File "/data/sol.prod/w
ww/payments.firefox.com/current/venv/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrap
ped_view#012    return view_func(*args, **kwargs)#012  File "/data/sol.prod/www/payments.firefox.com/current/venv/l
ib/python2.7/site-packages/rest_framework/views.py", line 399, in dispatch#012    response = self.handle_exception(
exc)#012  File "/data/sol.prod/www/payments.firefox.com/current/venv/lib/python2.7/site-packages/rest_framework/vie
ws.py", line 364, in handle_exception#012    response = self.settings.EXCEPTION_HANDLER(exc)#012  File "./solitude/
exceptions.py", line 16, in custom_exception_handler#012    set_rollback(True)#012  File "/data/sol.prod/www/paymen
ts.firefox.com/current/venv/lib/python2.7/site-packages/django/db/transaction.py", line 103, in set_rollback#012   
 return get_connection(using).set_rollback(rollback)#012  File "/data/sol.prod/www/payments.firefox.com/current/ven
v/lib/python2.7/site-packages/django/db/backends/base/base.py", line 313, in set_rollback#012    "The rollback flag
 doesn't work outside of an 'atomic' block.")#012TransactionManagementError: The rollback flag doesn't work outside
 of an 'atomic' block.
```

There's a couple of things to investigate here:
1) what is causing the failure in the first place
2) why isn't the roll back working

A suggested first change is to log the heck out of the problem as suggested by kumar303 here:

https://github.com/mozilla/solitude/blob/master/solitude/exceptions.py#L16

andymckay:
* fixes #520 
* log before we attempt the roll back
* `DATABASES['default']` is being overridden, meaning we weren't in an atomic block although the code assumed that

### DIFF
--- a/solitude/exceptions.py
+++ b/solitude/exceptions.py
@@ -13,6 +13,7 @@ log = getLogger('s')
 def custom_exception_handler(exc):
     # If you raise an error in solitude, it comes to here and
     # we rollback the transaction.
+    log.exception('Handling exception, about to roll back', exc_info=exc)
     set_rollback(True)
 
     if hasattr(exc, 'formatter'):
--- a/solitude/settings/sites/paymentsalt/db.py
+++ b/solitude/settings/sites/paymentsalt/db.py
@@ -17,6 +17,7 @@ DATABASES = {}
 DATABASES['default'] = dj_database_url.parse(private.DATABASES_DEFAULT_URL)
 DATABASES['default']['ENGINE'] = 'django.db.backends.mysql'
 DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
+DATABASES['default']['ATOMIC_REQUESTS'] = True
 
 DEBUG = False
 DEBUG_PROPAGATE_EXCEPTIONS = False
--- a/solitude/settings/sites/dev/db.py
+++ b/solitude/settings/sites/dev/db.py
@@ -21,6 +21,7 @@ DATABASES = {}
 DATABASES['default'] = dj_database_url.parse(private.DATABASES_DEFAULT_URL)
 DATABASES['default']['ENGINE'] = 'django.db.backends.mysql'
 DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
+DATABASES['default']['ATOMIC_REQUESTS'] = True
 
 DEBUG = False
 DEBUG_PROPAGATE_EXCEPTIONS = False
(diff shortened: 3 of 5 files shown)


solitude/exceptions.py
```diff
@@ -13,6 +13,7 @@
 def custom_exception_handler(exc):
     # If you raise an error in solitude, it comes to here and
     # we rollback the transaction.
+    log.exception('Handling exception, about to roll back')
```
kumar303: I'm not sure that it's safe to do this because even though it logs the last exception in the system stack, Django might caught something innocuous (like a KeyError) already.

kumar303: I think you should log `exc` directly

solitude/settings/sites/paymentsalt/db.py
```diff
@@ -17,6 +17,7 @@
 DATABASES['default'] = dj_database_url.parse(private.DATABASES_DEFAULT_URL)
 DATABASES['default']['ENGINE'] = 'django.db.backends.mysql'
 DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
+DATABASES['default']['ATOMIC_REQUESTS'] = True
```
kumar303: hmm, this isn't very DRY. I suppose it will do for now. Seems like `DATABASES['default'] = base.DATABASES['default'].copy()` should be possible though.

kumar303:
r+wc

solitude/exceptions.py
andymckay: That seems legit, will do.

solitude/settings/sites/paymentsalt/db.py
kumar303: or maybe make a `base.DEFAULT_DB_SETTINGS` dict that everything can use because otherwise we'll probably forget and introduce another bug like this in the future.

PR #524, opened 2015-08-07
fix the ordering

### LINKED ISSUE #523: DRF not respecting model ordering
There's a default ordering set on the model of -created. Turns out django-filter is stripping that and no order is being applied, so responses from solitude are unordered.

andymckay:
* fixes #523 
* pass on the ordering on the model down to django-filter, kinda confused why DRF doesn't do this already
* can't think of anything that explicitly relies on ordering in solitude, so should be ok

fyi @muffinresearch, this will ensure payment methods are ordered newest first

### DIFF
--- a/solitude/filter.py
+++ b/solitude/filter.py
@@ -1,6 +1,9 @@
 from rest_framework.filters import DjangoFilterBackend
 
 from solitude.errors import InvalidQueryParams
+from solitude.logger import getLogger
+
+log = getLogger('s.filter')
 
 
 class StrictQueryFilter(DjangoFilterBackend):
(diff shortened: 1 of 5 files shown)


solitude/filter.py
```diff
@@ -10,6 +13,16 @@ class StrictQueryFilter(DjangoFilterBackend):
     Instead limit it down to the parameters allowed in filter_fields.
     """
 
+    def get_filter_class(self, view, queryset=None):
+        klass = (super(StrictQueryFilter, self)
+                 .get_filter_class(view, queryset=queryset))
+        try:
+            # If an ordering exists on the model, use that.
+            klass._meta.order_by = klass.Meta.model.Meta.ordering
```
kumar303: could this just be `hasattr(klass.Meta.model.Meta, 'ordering')` ? Or could it be missing `.model` too?

kumar303:
I don't know the specifics but in addition to the workaround, would it make sense to submit a bug upstream?

r+wc

solitude/filter.py
andymckay: not sure, I think `klass.Meta.model.Meta` should always exist, this is just future proofing in case someone doesn't inherit from solitude.base

andymckay:
I would rather see how DRF 3.x behaves before filing a bug, wasn't sure if this was just being behind on DRF or not.

PR #525, opened 2015-08-07
always migrate

andymckay:
* this is pretty fast when its a no-op and will ensure that on database updates, things just don't blow up


muffinresearch:
R+

PR #528, opened 2015-08-11
Filter transactions by user

kumar303:
Supports https://github.com/mozilla/payments-ui/issues/279

### DIFF
--- a/lib/brains/tests/test_transaction.py
+++ b/lib/brains/tests/test_transaction.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from urllib import urlencode
 
 from django.core.urlresolvers import reverse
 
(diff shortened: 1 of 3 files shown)


muffinresearch:
lgtm r+

lib/brains/tests/test_transaction.py
```diff
@@ -19,14 +20,15 @@ def setUp(self):
...
+                                                      buyer=self.buyer)
+        super(TestBraintreeTransaction, self).setUp()
 
     def test_allowed(self):
         self.allowed_verbs(self.url, ['get'])
 
-    def create(self):
-        return BraintreeTransaction.objects.create(
+    def create(self, **attributes):
+        final_attributes = dict(
```
andymckay: generally try to avoid dict unless there's a good reason for it

PR #529, opened 2015-08-12
api_name was a tastypie-ism

andymckay:


kumar303:
I like getting rid of tastypie-isms, r+

PR #530, opened 2015-08-12
close account api

### LINKED ISSUE #497: Support delete account

andymckay:
* hook up model method into the API
* `/generic/buyer/id/close/` is a nice URL, but kinda awkward to put together in payments-service, maybe? if you want a different URL kumar let me know.
* doesn't return anything about the closing... dunno if you want anything
* fixes #497

### DIFF
--- a/docs/topics/generic.rst
+++ b/docs/topics/generic.rst
@@ -213,7 +213,7 @@ After these two steps you will use the ``reset_confirm_pin`` endpoint. It works
 the same way as the ``confirm_pin`` endpoint but instead checks against the
 buyer's ``new_pin`` rather than their ``pin``:
 
-.. htp:post::: /generic/reset_confirm_pin/
+.. http:post::: /generic/reset_confirm_pin/
 
     **Request**
 
@@ -246,6 +246,42 @@ minutes. You can tell if a buyer is locked by checking the
 ``pin_is_locked_out`` property of the buyer data. Buyers that were locked out
 since the last time the PIN was changed or successfully verified will have the
 ``pin_was_locked_out`` property set to ``true``.
+
+Close
+-----
+
+This does not delete the buyer, but does the following:
+
+* calls a close signal the payment providers can listen to, in the case of Braintree it cancels all payment methods and subscriptions
+* sets the account to inactive
+* removes the email
+* sets the uuid to something anonymous
+
+Note: only if the close signal is succesfully processed will the account be set to inactive.
+
+.. http:post:: /generic/buyer/int:id/close/
+
+    :status 204: account closed successfully
+    :status 400: problem processing the uuid into a buyer
+    :status 404: active buyer not found, will trigger if you try to close an account twice
+    :status 500: something went wrong with closing the account
+
+The account is not deleted and can still be accessed by the URL to that account to preserve data integrity. For example:
+
+* create a buyer with a POST to `/generic/buyer`
+* store the `resource_uri` in the response
+* close the buyer with a POST to `/generic/buyer/int:id/close`
+* get the buyer with a GET to `resource_uri`, a *truncated* version of the response shows:
+
+  .. code-block:: json
+
+    {
+        "active": False,
+        "email": "",
+        "uuid": "anonymised-uuid:1b3db7a9-0e8f-43d8-b8da-b3317a147068"
+    }
+
+
 
 .. _seller-label:
 
--- a/lib/buyers/tests/test_api.py
+++ b/lib/buyers/tests/test_api.py
@@ -374,3 +374,22 @@ class TestBuyerResetPin(APITest):
     def test_empty_post(self):
         res = self.client.post(self.list_url, data={})
         eq_(res.status_code, 422)
+
+
+class TestBuyerClose(APITest):
+
+    def setUp(self):
+        self.uuid = 'sample:uid'
+        self.buyer = Buyer.objects.create(uuid=self.uuid)
+        self.url = reverse('generic:close', kwargs={'pk': self.buyer.pk})
+
+    def test_close(self):
+        res = self.client.post(self.url)
+        eq_(res.status_code, 204)
+        eq_(self.buyer.reget().active, False, res.content)
+
+    def test_already_closed(self):
+        self.buyer.active = False
+        self.buyer.save()
+        res = self.client.post(self.url)
+        eq_(res.status_code, 404, res.content)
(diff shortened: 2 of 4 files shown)


docs/topics/generic.rst
```diff
@@ -247,6 +247,42 @@ minutes. You can tell if a buyer is locked by checking the
...
+
+    :status 200: account closed successfully
+    :status 400: problem processing the uuid into a buyer
+    :status 404: active buyer not found, will trigger if you try to close an account twice
+    :status 500: something went wrong with closing the account
+
+The account is not deleted and can still be accessed by the URL to that account to preserve data integrity. For example:
+
+* create a buyer with a POST to `/generic/buyer`
+* store the `resource_uri` in the response
```
andymckay: should be "...from the response"

docs/topics/generic.rst
```diff
@@ -213,7 +213,7 @@ After these two steps you will use the ``reset_confirm_pin`` endpoint. It works
 the same way as the ``confirm_pin`` endpoint but instead checks against the
 buyer's ``new_pin`` rather than their ``pin``:
 
-.. htp:post::: /generic/reset_confirm_pin/
+.. http:post::: /generic/reset_confirm_pin/
```
kumar303: heh, oops. We should really start building docs as part of the test suite (to check for errors).

docs/topics/generic.rst
```diff
@@ -247,6 +247,42 @@ minutes. You can tell if a buyer is locked by checking the
...
+    :status 200: account closed successfully
+    :status 400: problem processing the uuid into a buyer
+    :status 404: active buyer not found, will trigger if you try to close an account twice
+    :status 500: something went wrong with closing the account
+
+The account is not deleted and can still be accessed by the URL to that account to preserve data integrity. For example:
+
+* create a buyer with a POST to `/generic/buyer`
+* store the `resource_uri` in the response
+* close the buyer with a POST to `/generic/buyer/int:id/close`
```
kumar303: might want to put a trailing slash on this URL just for completeness

docs/topics/generic.rst
```diff
@@ -247,6 +247,42 @@ minutes. You can tell if a buyer is locked by checking the
...
+  .. code-block:: json
+
+    {
+        "active": False,
+        "email": "",
+        "uuid": "anonymised-uuid:1b3db7a9-0e8f-43d8-b8da-b3317a147068"
+    }
+
+
+
```
kumar303: nice docs

docs/topics/generic.rst
```diff
@@ -247,6 +247,42 @@ minutes. You can tell if a buyer is locked by checking the
...
+* calls a close signal the payment providers can listen to, in the case of Braintree it cancels all payment methods and subscriptions
+* sets the account to inactive
+* removes the email
+* sets the uuid to something anonymous
+
+Note: only if the close signal is succesfully processed will the account be set to inactive.
+
+.. http:post:: /generic/buyer/int:id/close/
+
+    :status 200: account closed successfully
```
kumar303: I think this changed to 204?

lib/buyers/tests/test_api.py
```diff
@@ -374,3 +374,22 @@ def test_bad_uuid(self):
...
+
+    def setUp(self):
+        self.uuid = 'sample:uid'
+        self.buyer = Buyer.objects.create(uuid=self.uuid)
+        self.url = reverse('generic:buyer-close', kwargs={'pk': self.buyer.pk})
+
+    def test_close(self):
+        res = self.client.post(self.url, data={'uuid': self.uuid})
+        eq_(res.status_code, 204)
+        eq_(self.buyer.reget().active, False, res.content)
```
kumar303: I think it's worth testing that the email is blank and that the uuid is different

lib/buyers/tests/test_api.py
```diff
@@ -374,3 +374,22 @@ def test_bad_uuid(self):
...
+
+class TestBuyerClose(APITest):
+
+    def setUp(self):
+        self.uuid = 'sample:uid'
+        self.buyer = Buyer.objects.create(uuid=self.uuid)
+        self.url = reverse('generic:buyer-close', kwargs={'pk': self.buyer.pk})
+
+    def test_close(self):
+        res = self.client.post(self.url, data={'uuid': self.uuid})
```
kumar303: this doesn't need to post a uuid, does it?

lib/buyers/views.py
```diff
@@ -14,6 +19,13 @@ class BuyerViewSet(NonDeleteModelViewSet):
     serializer_class = BuyerSerializer
     filter_fields = ('uuid', 'active')
 
+    @detail_route(methods=['post'])
```
kumar303: this is shiny but can't you just do : 
````
class CloserBuyer(APIView):
    def post(self, request, pk):
        ...
url(r'^generic/buyer/(?P<pk>[\d]+)/close/$', CloserBuyer.as_view())
````
?

kumar303:
The URL `/generic/buyer/:id/close/` seems fine, we can hook it up in the service easily with regex

kumar303:
r+wc

docs/topics/generic.rst
andymckay: oh right it is, ta

lib/buyers/tests/test_api.py
andymckay: that's in the tests for the close() method: https://github.com/mozilla/solitude/commit/2fd58fedf5fb3af517141248acca43f4ad8bdca7#diff-ba0f8932616d767f6d87927a25578a6eR120

lib/buyers/tests/test_api.py
andymckay: it does not, that was my original plan

lib/buyers/views.py
andymckay: I could. Since the lookup is doing get_object_or_404 theres no real advantage in being in the BuyerViewSet... so only real advantage is that @detail_route is in DRF 3, so when we get upgrade (and delete the @detail_route), this ends up being slightly less code.

kumar303: the only issue I have with `detail_route` is you didn't copy over tests which means it would be hard to upgrade if we need an intermediary release before 3. Hopefully we can go straight to 3 though.

PR #535, opened 2015-08-17
add in transaction.sale api

### LINKED ISSUE #532: Implement transaction sale for one-off payments
Allow a one-off payment by implementing this API:

https://developers.braintreepayments.com/javascript+python/reference/request/transaction/sale

The solitude API should probably accept a product type so we can tie it back to a product on solitudes end.

Tracking mozilla/payments#117

andymckay:
* fixes #532

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -287,6 +287,77 @@ Some information about the subscripton is stored in solitude.
     :query provider_id: the plan id for this subscription.
     :query seller_product: the primary key of the product.
 
+Sale
+----
+
+A sale is a one off payment to call the Braintree Transaction API. For more
+information see the `Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/request/transaction/sale>`_.
+
+This should not be used for subscriptions.
+
+.. http:post:: /braintree/sale/
+
+    :<json amount: the amount of the transaction, within the maximum and minimum limits
+    :<json product_id: the product_id as defined by `payments-config <https://github.com/mozilla/payments-config/>`_.
+    :<json nonce: (optional) the payment nonce returned by Braintree, used when no payment method is stored.
+    :<json paymethod: (optional) the URI to `a payment object <payment-methods-label>`_.
+
+    .. code-block:: json
+
+        {
+          "mozilla": {
+            "generic": {
+              "resource_pk": 1,
+              "related": null,
+              "seller_product": "/generic/product/1/",
+              "currency": "USD",
+              "uid_pay": null,
+              "uuid": "",
+              "uid_support": "test-id",
+              "relations": [],
+              "seller": "/generic/seller/1/",
+              "source": null,
+              "provider": 4,
+              "pay_url": null,
+              "type": 0,
+              "status": 2,
+              "buyer": null,
+              "status_reason": null,
+              "created": "2015-08-17T19:17:04.296",
+              "notes": null,
+              "amount": "5.00",
+              "carrier": null,
+              "region": null,
+              "resource_uri": "/generic/transaction/1/"
+            },
+            "braintree": {
+              "kind": "",
+              "transaction": "/generic/transaction/1/",
+              "next_billing_period_amount": null,
+              "created": "2015-08-17T19:17:04.298",
+              "paymethod": null,
+              "counter": 0,
+              "billing_period_end_date": null,
+              "modified": "2015-08-17T19:17:04.298",
+              "next_billing_date": null,
+              "resource_pk": 1,
+              "resource_uri": "/braintree/mozilla/transaction/1/",
+              "billing_period_start_date": null,
+              "id": 1,
+              "subscription": null
+            }
+          },
+          "braintree": {}
+        }
+
+
+    :>json mozilla.generic: the generic transaction object.
+    :>json mozilla.generic.buyer: this will be the buyer or empty if no buyer is registered.
+    :>json mozilla.braintree: the braintree transaction object.
+
+Notes:
+* either a `nonce` or a `paymethod` must exist, but not both
+
 Webhook
 -------
 
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -1,3 +1,5 @@
+from decimal import Decimal
+
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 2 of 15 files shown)


docs/topics/braintree.rst
```diff
@@ -287,6 +287,77 @@ Some information about the subscripton is stored in solitude.
     :query provider_id: the plan id for this subscription.
     :query seller_product: the primary key of the product.
 
+Sale
+----
+
+A sale is a one off payment to call the Braintree Transaction API. For more
+information see the `Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/request/transaction/sale>`_.
+
+Cannot be used for subscriptions.
```
kumar303: nit: "This cannot ...."

lib/brains/forms.py
```diff
@@ -188,6 +190,75 @@ def clean(self):
         return self.cleaned_data
 
 
+class SaleForm(forms.Form):
+    amount = forms.DecimalField(
+        # Arbitrary values should they be set in a config somewhere?
```
andymckay: i guess i did set them in a config :)

lib/brains/forms.py
```diff
@@ -188,6 +190,75 @@ def clean(self):
...
+        min_value=Decimal(settings.BRAINTREE_MIN_AMOUNT))
+    # This will be populated by the paymethod.
+    buyer = None
+    nonce = forms.CharField(max_length=255, required=False)
+    paymethod = PathRelatedFormField(
+        view_name='braintree:mozilla:paymethod-detail',
+        queryset=BraintreePaymentMethod.objects.filter(),
+        required=False,
+        allow_null=True)
+    # Product, seller and seller_product are set by looking up the
```
andymckay: i removed product, comment needs updating

lib/brains/forms.py
```diff
@@ -188,6 +190,75 @@ def clean(self):
...
+
+    def clean_product_id(self):
+        product_id = self.cleaned_data.get('product_id')
+
+        try:
+            self.seller_product = (
+                SellerProduct.objects.get(public_id=product_id))
+        except SellerProduct.DoesNotExist:
+            raise forms.ValidationError(
+
```
andymckay: stupid new line

docs/topics/braintree.rst
```diff
@@ -287,6 +287,77 @@ Some information about the subscripton is stored in solitude.
...
+
+.. http:post:: /braintree/sale/
+
+    :<json amount: the amount of the transaction, within the maximum and minimum limits
+    :<json product_id: the product_id as defined by `payments-config <https://github.com/mozilla/payments-config/>`_
+    :<json nonce: (optional) product nonce, for use where no payment method is stored
```
kumar303: there are a lot of different nonces - can we call this `paymethod_nonce` ? Also, you call it a 'product nonce' but I don't think that's correct.

docs/topics/braintree.rst
```diff
@@ -287,6 +287,77 @@ Some information about the subscripton is stored in solitude.
...
+    :<json paymethod: (optional) the URI to `a payment object <payment-methods-label>`_.
```
kumar303: I know we have this pattern in a few places but I think it would be helpful to name this more specifically like `paymethod_uri`. That way you can glance at it and know what it is.

lib/brains/forms.py
kumar303: nit: extra space

lib/brains/models.py
```diff
@@ -140,16 +140,18 @@ class BraintreeTransaction(Model):
     some of this is not stored in the generic transaction.
     """
     # There isn't enough information on the Transaction.
```
kumar303: I don't remember what this comment meant. Does it still apply?

lib/brains/tests/test_forms.py
```diff
@@ -84,3 +85,58 @@ def test_missing_cancel_params(self):
...
+
+    def test_both_paymethod_nonce(self):
+        method = create_method(create_braintree_buyer()[1])
+        form, errors = self.errors({
+            'nonce': 'abc',
+            'paymethod': method.get_uri()
+        })
+        assert '__all__' in errors, errors.as_text()
+
+    def test_buyer_non_existant(self):
```
kumar303: `test_paymethod_non_existent` ?

docs/topics/braintree.rst
andymckay: I don't think we have any more nonces, we have lots of tokens... its called a nonce here http://solitude.readthedocs.org/en/latest/topics/braintree.html#post--braintree-paymethod- and its the same value.

docs/topics/braintree.rst
andymckay: I'd vote for consistency its called paymethod in at least 4 other APIs e.g. http://solitude.readthedocs.org/en/latest/topics/braintree.html#post--braintree-subscription-

lib/brains/tests/test_forms.py
```diff
@@ -84,3 +85,58 @@ def test_missing_cancel_params(self):
         errors = form.errors
         # Make sure empty params are caught.
         assert 'subscription' in errors, errors.as_text()
+
+
+class TestSaleForm(BraintreeTest):
+
+    def errors(self, data):
```
kumar303: the calling code reads oddly when this is called `errors()`. Maybe just `submit()` ?

lib/brains/models.py
andymckay: no, was a comment as i was developing

lib/brains/views/sale.py
```diff
@@ -0,0 +1,52 @@
...
+        buyer=form.buyer,
+        currency=result.transaction.currency_iso_code,
+        provider=constants.PROVIDER_BRAINTREE,
+        seller=form.seller_product.seller,
+        seller_product=form.seller_product,
+        status=constants.STATUS_CHECKED,
+        type=constants.TYPE_PAYMENT,
+        uid_support=result.transaction.id
+    )
+    braintree_transaction = BraintreeTransaction.objects.create(
```
kumar303: if `form.paymethod` is specified, we need to pass it in to `create()` I think

kumar303: how does the braintree transaction `kind` get set? Shouldn't we set that here since we asked to submit for settlement immediate?

solitude/settings/base.py
```diff
@@ -510,3 +510,7 @@
 
 # An arbitrary limit on the number of payment methods.
 BRAINTREE_MAX_METHODS = 5
+
+# Arbitrary amounts for top and bottom limits.
+BRAINTREE_MAX_AMOUNT = 10000
+BRAINTREE_MIN_AMOUNT = 1
```
kumar303: Do we need to specify a min/max? I suspect these defaults are not quite right, especially on the minimum

lib/brains/forms.py
kumar303: yes, I think the product should be configured with something like `allow_variable_amount=True` and this form should bail if that's not True when accepting a declared amount. Could be done in a follow-up patch. Otherwise we could some `0.01` purchases of fixed price products ;)

solitude/settings/base.py
andymckay: A minimum definitely, otherwise Braintree will charge us more than we donate if its below about 30 cents. A maximum? Dunno.

lib/brains/views/sale.py
andymckay: Good call, if paymethod is there, I should add it into the transaction, will add.

docs/topics/braintree.rst
kumar303: it seems pretty vague to call it a nonce because even in the context of a sale it's not obvious that it represents a payment method. Just a nit though. I often am annoyed by things like this when I look at some calling code and need to read the docs when a more specific variable name would avoid a doc lookup.

docs/topics/braintree.rst
kumar303: yeah, I was bothered by those in my Rs too IIRC :) Sure, consistency wins I guess.

kumar303:
r+wc

lib/brains/views/sale.py
andymckay: Can't think of anything better to put in kind, so why not.

PR #539, opened 2015-08-19
Fix subscription_id in braintree_transaction table

kumar303:
This column was changed accidentally in
https://github.com/mozilla/solitude/pull/535


andymckay:
r+ thanks for fixing up my mistakes

PR #540, opened 2015-08-19
check the product config before some transaction.sale action

### LINKED ISSUE #537: The sale API should not allow variable amounts for all products
The sale API added in https://github.com/mozilla/solitude/pull/535 lets you pay for the product with a variable amount. This is great for donations but not great for products that have a fixed price :) We need to lock this down before we offer any non-donation products. Solitude can probably figure this out by checking if a product has `amount` in its configuration or not. For donations, no amount is configured.

andymckay:
* fixes #537
* also check the product is not recurring
* ended up rabbit holing on how to mock out product config
* could not get the mock to stick with the import at the top, which was frustrating

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -4,13 +4,14 @@ from django import forms
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 
+import payments_config
 import requests
 
 from lib.brains.models import (
     BraintreeBuyer, BraintreePaymentMethod, BraintreeSubscription)
 from lib.buyers.models import Buyer
 from lib.sellers.models import SellerProduct
-from payments_config import products
+
 from solitude.base import getLogger
 from solitude.related_fields import PathRelatedFormField
 
@@ -110,7 +111,7 @@ class SubscriptionForm(forms.Form):
         return 'Mozilla*{}'.format(name)[0:22]
 
     def get_name(self, plan_id):
-        if plan_id in products:
+        if payments_config.products.get(plan_id):
             return unicode(products.get(plan_id).description)
         log.warning('Unknown product for descriptor: {}'.format(plan_id))
         return 'Product'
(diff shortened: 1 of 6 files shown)


lib/brains/forms.py
```diff
@@ -110,7 +110,10 @@ def format_descriptor(self, name):
         return 'Mozilla*{}'.format(name)[0:22]
 
     def get_name(self, plan_id):
-        if plan_id in products:
+        # The only way I can get the mock to stick is to do a late import.
+        # If we can fix that we can move this up.
+        from payments_config import products
```
kumar303: did you try `import payments_config.products` at the top and then patching `payments_config` ? That should work. Not sure if it would be any prettier though. You could also try patching `get()` on `products` directly which would probably work as well.

lib/brains/forms.py
```diff
@@ -238,6 +241,35 @@ def clean(self):
...
+                .format(self.cleaned_data.get('product_id')),
+                code='invalid')
+
+        if (product.amount
+                and product.amount != self.cleaned_data.get('amount')):
+            self.add_error(
+                'amount',
+                forms.ValidationError(
+                    'Product has an amount specified: {} and the amount '
+                    'given differs: {}'.format(
```
kumar303: This reads awkwardly to me, how about: "This product can only be bought for {}; amount requested: {}"

lib/brains/forms.py
```diff
@@ -238,6 +241,35 @@ def clean(self):
...
+        product = products.get(self.cleaned_data.get('product_id'))
+
+        if not product:
+            raise forms.ValidationError(
+                'Product does not exist: {}'
+                .format(self.cleaned_data.get('product_id')),
+                code='invalid')
+
+        if (product.amount
+                and product.amount != self.cleaned_data.get('amount')):
```
kumar303: I think this is not quite right. It should be wrapped like this because amount should be optional:
````
if self.cleaned_data.get('amount'):
    if product.amount and product.amount != self.cleaned_data.get('amount'):
        ...
````

And then `amount` should be optional in the form because it really doesn't make sense to post an amount for fixed price products.

File: lib/brains/tests/test_forms.py [hunk 4]
```diff
@@ -131,12 +134,32 @@ def test_ok(self):
...
+            'amount': '10.00',
             'paymethod': method.get_uri(),
             'product_id': seller_product.public_id
         })
         assert not errors, errors.as_text()
         assert 'payment_method_nonce' not in form.braintree_data
         eq_(form.braintree_data['payment_method_token'], method.provider_id)
+
+    def test_recurring(self):
+        seller, seller_product = create_seller()
```
kumar303: this would be less confusing to me if you explicitly added the line `seller_product.recurrence = 'monthly'`

lib/brains/tests/test_forms.py
```diff
@@ -131,12 +134,32 @@ def test_ok(self):
...
+        form, errors = self.process({
+            'amount': '10.00',
+            'nonce': 'nonce',
+            'product_id': seller_product.public_id
+        })
+        eq_(form.errors.as_data()['product_id'][0].code, 'invalid')
+
+    def test_amount(self):
```
kumar303: why is this amount invalid? Just a guess but maybe the test should be named `test_fail_if_amount_too_low` ?

kumar303:
main thing is I think amount should be optional otherwise r+wc

lib/brains/forms.py
andymckay: I'll give it a try, was frustrating.

lib/brains/forms.py
andymckay: So my logic will fail when `amount` is not specified in the form, but is specified on the product. Yeah I see, will fix.

lib/brains/tests/test_forms.py
andymckay: because i'm trying to specify a different amount for a fixed price product, something we don't to allow, i'll change the title

PR #543, opened 2015-08-21
Allow variable amounts for donations (bug fix)

kumar303:
This wasn't checking configured product amounts correctly.

### DIFF
--- a/lib/brains/forms.py
+++ b/lib/brains/forms.py
@@ -249,7 +249,7 @@ class SaleForm(forms.Form):
                 code='invalid')
 
         amount = self.cleaned_data.get('amount')
-        if (amount and product.amount != amount):
+        if (product.amount and product.amount != amount):
             self.add_error(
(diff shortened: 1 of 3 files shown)


lib/brains/forms.py
```diff
@@ -249,7 +249,7 @@ def clean(self):
                 code='invalid')
 
         amount = self.cleaned_data.get('amount')
-        if (amount and product.amount != amount):
+        if (product.amount and product.amount != amount):
```
andymckay: you just changed it back to what I had the first time :)

andymckay:
r+ because it means i was right the first time.

lib/brains/forms.py
kumar303: well why did you change it? :fried_shrimp:

andymckay: So confused :confetti_ball: :whale2:

Issue #546, opened 2015-08-21
Allow subscriptions to have variable amounts


andymckay-limited-access:
Currently subscriptions are at fixed amounts. But we can create a subscription with a variable amount.

The data about the new price should probably be stored on the braintree subscription object so we can figure out what that price is later.

Docs: https://developers.braintreepayments.com/javascript+python/guides/recurring-billing/create#overriding-plan-details

The product configuration will need to allow this.

kumar303:
requires https://github.com/mozilla/solitude/issues/547

PR #548, opened 2015-08-24
Allow custom amount on braintree subscriptions (#546)

kumar303:
Fixes https://github.com/mozilla/solitude/issues/546

### DIFF
--- a/lib/brains/tests/test_forms.py
+++ b/lib/brains/tests/test_forms.py
@@ -1,5 +1,6 @@
 from django.test.utils import override_settings
 
+from braintree.subscription_gateway import SubscriptionGateway
 from nose.tools import eq_
 
 from lib.brains.forms import (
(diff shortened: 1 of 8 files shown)


lib/brains/tests/test_forms.py
```diff
@@ -51,23 +54,77 @@ def test_parse_ok(self):
...
+        assert 'plan' in form.errors, form.errors.as_text()
+
+    def test_plan_amount_is_not_customizable(self):
+        form = self.submit(expect_errors=True, data={
+            # Link to a non-donation product.
+            'plan': 'moz-brick',
+            # You cannot change the price on a product with a fixed amount.
+            'amount': '1.99',
+        })
+        assert 'amount' in form.errors, form.errors.as_text()
```
muffinresearch: Since there's more than one error for amount could these tests check the code?

muffinresearch:
I think it might be nice to make the error assertions a bit more specific where possible. Just helps ensure the right error is happening at the right place and will also help with the future when more error cases are added under the same key.

r+wc

lib/brains/tests/test_forms.py
kumar303: oh yeah, good call. I don't know if I can get at the code but I can pattern match part of the message

kumar303: done

PR #549, opened 2015-08-25
Mark buyers as authenticated or not (fixes #545)

### LINKED ISSUE #545: Create a subscription using a non-authenticated buyer
This will support whatever mozilla/payments-service#141 needs in the backend. From a solitude point of view, we create all the usual subscription objects except the buyer object will only have an email address, i.e. no FxA data. 

Since FxA matches up with the buyer's uuid, it might just be a matter of creating a uuid that is prefixed with "UNVERIFIED:" or something. We do similar things for anonymisation.

It might be worth blocking listing payment methods for an unverified users to be sure payments-service doesn't do anything naughty... or maybe that sort of logic should be in payments-service.

Supports: https://github.com/mozilla/payments/issues/117

kumar303:
Fixes https://github.com/mozilla/solitude/issues/545

This should allow us to create an email-only buyer (in services) before creating a recurring donation. Well, it was always possible to create an email-only buyer but this new flag will allow us to put restrictions elsewhere in Solitude if needed.

### DIFF
--- a/docs/topics/braintree.rst
+++ b/docs/topics/braintree.rst
@@ -25,7 +25,7 @@ customer exists in Braintree already, it is not created.
 
 .. http:post:: /braintree/customer/
 
-    :<json string uuid: the uuid of the buyer in solitude.
+    :<json string uuid: the UUID of the :ref:`buyer <buyer-label>`.
 
     .. code-block:: json
 
@@ -300,7 +300,7 @@ Sale
 ----
 
 A sale is a one off payment to call the Braintree Transaction API. For more
-information see the `Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/request/transaction/sale>`_.
+information see the `Braintree transaction documentation <https://developers.braintreepayments.com/javascript+python/reference/request/transaction/sale>`_.
 
 This should not be used for subscriptions.
 
@@ -373,7 +373,7 @@ Webhook
 When Braintree completes certain actions, they will make a request to the
 configured webhook URL. That will be `payments-service <https://github.com/mozilla/payments-service/>`_
 which then passes it on to this endpoint. For more information see the
-`Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks>`_.
+`Braintree webhook documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks/overview>`_.
 
 .. http:get:: /braintree/webhook/
 
--- a/docs/topics/generic.rst
+++ b/docs/topics/generic.rst
@@ -12,7 +12,7 @@ Buyers
 
 Buyers are identified by a UUID, which is a string (max 255 chars) that makes
 sense to the client. It must be unique within solitude, so we'd recommend
-prefixing the UUID, eg: `marketplace:<your-uuid>`
+prefixing the UUID, eg: ``marketplace:<your-uuid>``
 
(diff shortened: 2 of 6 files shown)


docs/topics/generic.rst
```diff
@@ -12,7 +12,7 @@ Buyers
 
 Buyers are identified by a UUID, which is a string (max 255 chars) that makes
 sense to the client. It must be unique within solitude, so we'd recommend
-prefixing the UUID, eg: `marketplace:<your-uuid>`
+prefixing the UUID, eg: ``marketplace:<your-uuid>``
```
muffinresearch: Why the double `?

lib/buyers/tests/test_api.py
```diff
@@ -29,6 +29,8 @@ def test_add(self):
         buyer = Buyer.objects.get(uuid=self.uuid)
         eq_(buyer.email, self.email)
         eq_(res.json['pin'], True)
+        # By default, we assume buyers have been authenticated.
+        eq_(res.json['authenticated'], True)
```
muffinresearch: What happens if this isn't the case?

docs/topics/braintree.rst
```diff
@@ -373,7 +373,7 @@ Webhook
 When Braintree completes certain actions, they will make a request to the
 configured webhook URL. That will be `payments-service <https://github.com/mozilla/payments-service/>`_
 which then passes it on to this endpoint. For more information see the
-`Braintree documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks>`_.
+`Braintree webhook documentation <https://developers.braintreepayments.com/javascript+python/reference/general/webhooks>`_.
```
muffinresearch: As an aside I noticed the braintree docs have moved some urls around so it's worth double-checking the links work. (we should probably automate that).

lib/buyers/tests/test_api.py
```diff
@@ -41,16 +43,26 @@ def test_add_multiple(self):
         eq_(self.get_errors(res.content, 'uuid'),
             [BUYER_UUID_ALREADY_EXISTS])
 
-    def test_add_empty(self):
+    def test_uuid_cannot_be_empty(self):
```
muffinresearch: Good call on the test func renaming - makes it clearer.

muffinresearch:
r+wc

docs/topics/generic.rst
kumar303: the single ticks make *emphasis* in restructured text, double ticks makes `code`. I think this was just a typo because confusingly single ticks in markdown will make code.

lib/buyers/tests/test_api.py
kumar303: Ha, true. I'll change 'we' to 'the code' ;)

docs/topics/braintree.rst
kumar303: mmm, there is a `linkcheck` command in sphinx actually. We could add it to the tests.

lib/buyers/tests/test_api.py
kumar303: I don't usually like to make extraneous changes in a patch but I whenever I read one of andy's test names and can't figure it out I spend time to figure it out then rename the test ;)

muffinresearch: :+1:  for random acts of kindness :)

PR #552, opened 2015-08-28
Configure recurring donations

kumar303:
This was needed for https://github.com/mozilla/payments-ui/issues/368

### DIFF
--- a/lib/brains/management/commands/braintree_config.py
+++ b/lib/brains/management/commands/braintree_config.py
@@ -1,4 +1,6 @@
+# -*- coding: utf-8 -*-
 from decimal import Decimal
+import textwrap
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
@@ -9,6 +11,12 @@ from payments_config import sellers
 from solitude.logger import getLogger
 
 log = getLogger('s.brains.management')
+
+
+class BraintreePlanDoesNotExist(Exception):
+    """
+    Failed to look up a recurring payment plan in the Braintree API.
+    """
 
 
 def get_or_create_seller(uuid):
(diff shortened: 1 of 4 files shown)


lib/brains/management/commands/braintree_config.py
```diff
@@ -100,4 +107,29 @@ def handle(self, *args, **options):
...
+                            2. Go to Recurring Billing > Plans from the sidebar
+                            3. Enter the following data:
+
+                                Plan ID: {product.id}
+                                Plan Name: {product.description}
+                                Price: {price}
+                                Billing Cycle Every: 1 month
+
+                            4. re-run this script
+                            5. MFBT
```
tofumatt: This is cute, but might look obscure to someone not indoctrinated into old Mozilla speak. I say add a link or make it a bit more generic. I wouldn't want someone to get confused when they're really at the end!

The :beer: emoji is an excellent alternative :wink:

lib/brains/management/commands/braintree_config.py
```diff
@@ -40,23 +48,23 @@ def product_exists(plans, external_id, amount):
...
     """
     if external_id not in plans:
-        log.warning(
-            'The plan: {0} does not exist in Braintree. Plans cannot be '
-            'created from the API and must be created from the Braintree '
-            'console.'
-            .format(external_id))
-
-        raise CommandError('Missing product: {0}'.format(external_id))
+        raise BraintreePlanDoesNotExist(
```
andymckay: +1 for more explicit errors, would move them into https://github.com/mozilla/solitude/blob/master/lib/brains/errors.py and import them though.

lib/brains/management/commands/braintree_config.py
```diff
@@ -100,4 +108,29 @@ def handle(self, *args, **options):
...
+                            2. Go to Recurring Billing > Plans from the sidebar
+                            3. Enter the following data:
+
+                                Plan ID: {product.id}
+                                Plan Name: {product.description}
+                                Price: {price}
+                                Billing Cycle Every: 1 month
+
+                            4. re-run this script
+                            5. MFBT 🍺
```
andymckay: Thanks for adding that.

andymckay:
r+

lib/brains/management/commands/braintree_config.py
kumar303: this is just for the config script (I doubt we'd need it anywhere else). It needed to raise an error so that the message could be displayed with more top-level data.

PR #553, opened 2015-09-04
add in an email hash

andymckay:
- [x] create an email hash column
- [x] make the tests go faster
- [x] make the lookup in DRF use this column
- [x] complete tests

supports mozilla/payments-service#151

### DIFF
--- a/docs/topics/generic.rst
+++ b/docs/topics/generic.rst
@@ -68,7 +68,24 @@ a unique UUID as well as the PIN the buyer has chosen:
 
     :status 201: successfully processed.
 
-You can also get the details of a buyer:
+List and retrieve
+-----------------
+
+List buyers:
+
+.. http:get:: /generic/buyer/
+
+    You can filter on the following parameters.
+
+    :param active: the active flag for a user.
+    :param email: users email address.
+    :param uuid: the uuid for a user.
+
+    **Response**
+
+    A standard listing response containing buyers (see below).
+
+Get the details of a buyer:
 
 .. http:get:: /generic/buyer/int:id/
 
--- a/lib/buyers/field.py
+++ b/lib/buyers/field.py
@@ -1,18 +1,20 @@
+import hashlib
+import hmac
+
+from django.conf import settings
 from django.contrib.auth.hashers import (check_password, get_hasher,
                                          make_password)
 from django.db.models import CharField, SubfieldBase
 
 
-class HashField(CharField):
+class HashFieldBase(CharField):
 
     __metaclass__ = SubfieldBase
 
     def __init__(self, max_length=255, *args, **kwargs):
         # Longer than the average hash, should generally be more than enough.
-        super(HashField, self).__init__(max_length=max_length, *args, **kwargs)
-
-    def get_prep_lookup(self):
-        raise NotImplemented('Looking up by hash is not supported')
+        super(HashFieldBase, self).__init__(
+            max_length=max_length, *args, **kwargs)
 
     def get_prep_value(self, value, prepared=False):
         if not prepared and value:
@@ -23,7 +25,7 @@ class HashField(CharField):
             # to_python() unless you are doing a .update() in which case
             # to_python()  is not called and the value is passed in raw.
             if not isinstance(value, HashedData):
-                value = make_password(value, salt=self.salt)
+                value = self._hash(value, salt=self.salt)
             self.salt = False  # dump salt after saving.
         return value
 
@@ -40,8 +42,42 @@ class HashField(CharField):
         # to save the hash, the same value we see is what is put in the DB.
         if not value.startswith(hasher.algorithm):
             self.salt = hasher.salt()
-            value = make_password(value, salt=self.salt)
+            value = self._hash(value, salt=self.salt)
         return HashedData(value)
+
+
+class HashField(HashFieldBase):
+
+    def get_prep_lookup(self, *args, **kw):
+        raise NotImplemented('Looking up by hash is not supported')
+
+    def _hash(self, value, salt=None):
+        return make_password(value, salt)
+
+
+class ConsistentSigField(HashFieldBase):
+
+    """
+    Like the HashField, but is always a consistent value to allow lookups
+    on the field.
+    """
+
+    def get_prep_lookup(self, *args, **kw):
+        return super(CharField, self).get_prep_lookup(*args, **kw)
+
+    def _hash(self, value, salt=None):
+        """
+        Hash the value with hmac, using sha256. This ignores values of
+        salt, so that the hash can be consistent.
+        """
+        if not value:
+            return ''
+        if value.startswith('consistent:'):
+            return value
+        key_id = max(settings.HMAC_KEYS.keys())
+        key_value = settings.HMAC_KEYS[key_id]
+        res = hmac.new(key_value, value, hashlib.sha256).hexdigest()
+        return 'consistent:' + res
 
 
 class HashedData(object):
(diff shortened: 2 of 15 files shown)


docs/topics/generic.rst
```diff
@@ -68,7 +68,24 @@ a unique UUID as well as the PIN the buyer has chosen:
 
     :status 201: successfully processed.
 
-You can also get the details of a buyer:
+List and retrieve
+-----------------
+
+List buyers:
+
+.. http:get:: /generic/buyer/?email=someone@somewhere.org
```
kumar303: does this need `?email=...` ?

docs/topics/generic.rst
```diff
@@ -68,7 +68,24 @@ a unique UUID as well as the PIN the buyer has chosen:
...
+
+    You can filter on the following parameters.
+
+    :param active: the active flag for a user.
+    :param email: users email address.
```
kumar303: user's

lib/buyers/field.py
```diff
@@ -23,7 +31,7 @@ def get_prep_value(self, value, prepared=False):
             # to_python() unless you are doing a .update() in which case
             # to_python()  is not called and the value is passed in raw.
             if not isinstance(value, HashedData):
-                value = make_password(value, salt=self.salt)
+                value = self._hash(value, salt=self.salt)
```
kumar303: can this just be `make_password(value, self.salt)` ?

kumar303: I see that this is what it was before, I just don't see why the instance method is needed.

kumar303: oh nevermind, that method is needed to implement the interface, I see now.

lib/buyers/field.py
```diff
@@ -40,10 +48,35 @@ def to_python(self, value):
...
+        return super(CharField, self).get_prep_lookup(*args, **kw)
+
+    def _hash(self, value, salt=None):
+        """
+        Hash the value with hmac, using sha256. This ignores values of
+        salt, so that the hash can be consistent.
+        """
+        if not value:
+            return ''
+        if value.startswith('consistent:'):
```
kumar303: this seems risky and makes me feel twitchy. Why is it needed? What if someone was able to fool the system by inputting a value like `consistent:whatever-else-they-want` ?

lib/buyers/field.py
```diff
@@ -40,10 +48,35 @@ def to_python(self, value):
...
 
+class ConsistentHashField(HashField):
+
+    """
+    Like the HashField, but is always a consistent value to allow lookups
+    on the field.
+    """
+
+    def get_prep_lookup(self, *args, **kw):
+        return super(CharField, self).get_prep_lookup(*args, **kw)
```
kumar303: this is an atypical use of super and I think it wouldn't be good to rely on this (another dev might delete this code because it looks like a mistake). If the `HashField` constructor is incompatible I suggest making an abstract `HashFieldBase` and inherit each one from that instead.

lib/buyers/models.py
```diff
@@ -100,9 +104,14 @@ def close(self):
         # All succeeds, so go ahead and anonymise the account.
         self.active = False
         self.email = ''
+        self.email_hash = None
```
kumar303: good idea!

lib/buyers/models.py
```diff
@@ -100,9 +104,14 @@ def close(self):
...
         self.uuid = ANONYMISED + str(uuid.uuid4())
         self.save()
         log.warning('Anonymising account complete: {}'.format(self.pk))
 
     def get_uri(self):
         return reverse('generic:buyer-detail', kwargs={'pk': self.pk})
+
+    def save(self, *args, **kw):
+        self.email_hash = self.email
```
kumar303: what is this doing? Does it hash the value later on in the chain? Can you add a comment? Otherwise it looks like a plain email is getting saved.

lib/buyers/models.py
```diff
@@ -27,6 +27,10 @@ class Buyer(Model):
     new_pin = HashField(blank=True, null=True)
     needs_pin_reset = models.BooleanField(default=False)
     email = AESField(blank=True, null=True, aes_key='buyeremail:key')
+    # Because the email field is encrypted we can't do lookups in mysql on it.
+    # This allows us to use a field for lookups, without exposing anything in
+    # the database.
+    email_hash = ConsistentHashField(blank=True, null=True)
```
kumar303: It may be pedantic so feel free to ignore this but I think this would be better named `email_sig` using `ConsistentSigField` because technically it is not a hash and if you were to actually hash the emails [Ben Adida would be sad](http://benlog.com/2008/06/19/dont-hash-secrets/). I am also suggesting this because I generally dislike naming things according to their implementation which is why I am not suggesting to name it `email_hmac` :)

lib/buyers/tests/test_api.py
```diff
@@ -63,18 +67,28 @@ def test_add_unauthenticated_buyer(self):
...
 
-    def create(self, **kwargs):
-        defaults = {'uuid': self.uuid, 'pin': self.pin, 'email': self.email,
-                    'locale': 'en-US'}
-        defaults.update(kwargs)
-        return Buyer.objects.create(**defaults)
+    def test_filter_email(self):
+        self.create()
+        res = self.client.get(self.list_url, {'email': self.email}).json
+        eq_(res['objects'][0]['email'], self.email)
```
kumar303: maybe assert length is 1 here too?

lib/buyers/tests/test_api.py
```diff
@@ -63,18 +67,28 @@ def test_add_unauthenticated_buyer(self):
...
+
+    def test_multiple_email(self):
+        self.create()
+        with self.assertRaises(ValueError):
+            self.client.get(self.list_url, {'email': [self.email, self.email]})
```
kumar303: shouldn't this return a 400? Or is the ValueError coming from django test tools?

kumar303: I see why. This seems like an odd side effect of using `HashedEmailRequest` and will cause a 500 for a request like `email=foo&email=bar...`. Did you consider using a form object or serializer instead?

kumar303: maybe a filter set would be easier? http://www.django-rest-framework.org/api-guide/filtering/#api-guide

solitude/settings/test.py
```diff
@@ -65,6 +65,11 @@
 STATIC_URL = '/'
 STATIC_ROOT = '/'
 
+# Not using bcrypt for hashing is a major test speed up.
```
kumar303: yay :+1: crypto on docker :-1:

kumar303:
I'm iffy about the `ValueError` and not thrilled about calling this a "hash" but I guess those aren't show stoppers. r+wc

docs/topics/generic.rst
andymckay: no, will remove

lib/buyers/field.py
andymckay: to prevent multiple hashing of the same value, you'll note HashField has the same code here: https://github.com/mozilla/solitude/blob/master/lib/buyers/field.py#L38-L41

lib/buyers/models.py
andymckay: I'll add a comment, when the email_hash gets saved to the db, the django Field hashes it.

lib/buyers/models.py
andymckay: ok

andymckay:
Didn't have time to clean up the `ValueError`, sorry about that. At least its explicit and will only manifest when a calling application is naughty. Shouldn't be a biggie.

lib/buyers/field.py
kumar303: luckily an email address can't contain `:` otherwise this would be a security problem ;) Hopefully the next thing that uses `ConsistentSigField` gets lucky too.

andymckay: suggestions on what to do would be helpful

andymckay: something like this might work https://github.com/mozilla/solitude/compare/mozilla:master...andymckay:move-sig-prefix?expand=1

PR #554, opened 2015-09-14
move prefix check

andymckay:
moves the check into `get_prep_value` which only occurs on de-serialization, not serialization, with a test to prove it


kumar303:
r+

Yeah, that's way better, thanks for the follow-up. I was going to suggest something like this but I wasn't sure if there was some other reason why it needed to be on `_hash`.

PR #555, opened 2015-11-25
bump for django security fix

andymckay:


kumar303:
I assume `pyrepo.stage.mozaws.net` will change soon to `pyrepo.mozaws.net` ? If so, could you file an issue to fix up the URL when ready?

r+

andymckay:
I don't know if the `stage` is going to be removed or not, @jasonthomas would know.

PR #556, opened 2016-02-05
sniff

andymckay:


tofumatt:
r+ :cry: :wink:

kumar303:
Good idea. I suppose webpay and spartacus needs one.

andymckay-limited-access:
I think that would be a nice heads up for contributors.

kumar303:
> contributors

they have contributors??

:ghost:

